PR #6, opened 2016-09-17
Improved usability

Rodjers:
Made the playbook idempotent by adding a creates clause to Install Consul and Install Consul UI. This prevents these tasks from failing when the playbook is rerun.

Changed node_name from invantory_hostname_short to inventory_hostname to prevent identical node_name when IP is used in the hosts file.

Added consul_group_name as a variable to enable users to use a different name for the group of consul nodes in their inventory by overriding this variable as cluster_nodes is very generic and may be used for other things.


Rodjers:
I'm not sure but it seems like the build is failing because of some error in how the tests are set up and not because if this change.

brianshumate:
Thanks for this; yes the test setup is causing the build failures so no worries.

PR #9, opened 2016-11-08
New option for DNS forwarding via iptables and recursors, small fixes:

crumohr:
- compatibility with Ubuntu 16.04
- booleans instead of strings for variables
- runnable if DNS is broken

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -15,7 +15,7 @@ consul_group_name: "cluster_nodes"
 consul_datacenter: "dc1"
 consul_domain: "consul"
 consul_log_level: "INFO"
-consul_syslog_enable: "true"
+consul_syslog_enable: true
 consul_iface: "eth1"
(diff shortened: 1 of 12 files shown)


brianshumate:
Thanks for this!

I will need to review a bit as it appears that the iptables tasks that you added are specific to Ubuntu only and do not have conditionals for RHEL based distros and this role should support both.

The quoted booleans are probably a holdover some old bug in a previous version of Ansible or elsewhere, so thanks for addressing that too!

I'll try to get this merged in soon.

brianshumate:
There's some great stuff in here. I see you made the iptables conditional so no worries and I guess CentOS and RHEL have it installed already so it'd just be skip for them.

Cool, I'll add you to CONTRIBUTORS as well.

Thanks again!

crumohr:
Thanks for merging and your comments!

I'll add a check to prevent the use of iptables on RHEL-based distros right away, on RHEL 7 and up it was replaced with [firewalld](https://fedoraproject.org/wiki/FirewallD).

PR #11, opened 2016-11-09
Allow iptables only on Debian-based distributions

crumohr:
- check for distribution compatibility
- allow override of `consul_raw_key`
- ACL setup in separate task


brianshumate:
👍 Nice idea to move ACLs to a separate task.

Thank you!

PR #12, opened 2016-11-09
Add dependencies for configuration by ansible

crumohr:
- Add client dependencies for further configuration by ansible (Debian-based only)
- Fix error using predefined encryption key
- Removal of redundant includes


brianshumate:
Looks good, thanks!

PR #17, opened 2016-11-16
Remove duplicated code

kostyrev:
Hi!
I'd like to propose to remove a lot of unnecessary code.
This will make the role simpler to read and understand.


brianshumate:
👍 This is good. I had added that stuff long ago in anticipation of any possible distribution specific binaries, but that will likely never happen so thanks for the cleanup!

PR #20, opened 2016-11-25
Allowing re-running to add new nodes. More HA too.

violuke:
Firstly, great role, it's really helped me, but it wasn't perfect in just a couple of ways, so I've made some changes to try to give back to the community. This PR does two things:

1. Previously if you re-ran the role with new nodes added, the new nodes would not be set up properly. They would not know the encryption key for a start, so could never connect.
2. The var `consul_bootstrap_address` was a bit too much of an assumption (your first node might not be the bootstrap node, but this assumed it was). Additionally, you should really be including the entire array of `server` nodes in the `start_join` config value, not just the first one. What if that node went down and then you restarted consul on another node? I've changed it to include all `server` and `bootstrap` nodes (apart from itself). Remember, the `bootstrap` node will become a `server` node shortly.
3. It does a similar with for the ACL token as it now does for encryption keys. Also the ACL replication token was never being set before, but now is.

Again, thanks for such a great role and I hope these changes help others too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,6 @@ inventory file (see below):
 | `consul_syslog_enable` | *true* | Log to syslog |
 | `consul_iface` | `eth1` | Consul network interface |
 | `consul_bind_address` | *127.0.0.1* | Bind address |
-| `consul_bootstrap_address` | `{{ hostvars[groups[consul_group_name][0]]['ansible_'+consul_iface]['ipv4']['address'] }}` | The server interface that additional server nodes will join to for bootstrapping |
 | `consul_dns_bind_address` | *127.0.0.1* | DNS API bind address |
 | `consul_http_bind_address` | *0.0.0.0* | HTTP API bind address |
 | `consul_https_bind_address` | *0.0.0.0* | HTTPS API bind address |
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -22,7 +22,6 @@ consul_dns_bind_address: "127.0.0.1"
 consul_http_bind_address: "0.0.0.0"
 consul_https_bind_address: "0.0.0.0"
 consul_rpc_bind_address: "0.0.0.0"
-consul_bootstrap_address: "{{ hostvars[groups[consul_group_name][0]]['ansible_'+consul_iface]['ipv4']['address'] }}"
 consul_node_name: "{{ inventory_hostname_short }}"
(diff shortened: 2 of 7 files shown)


violuke:
Edit: Added point 3 to first comment.

brianshumate:
Hi Luke,

Thanks for the PR!

I've been on a short holiday, but am back now and can review this and merge it soon.

Cheers,

Brian

violuke:
@bootswithdefer I have to say that I've not tested it with `--limit`, but I'd expect it to need at least 1 already setup consul node (can be a `server` or `client`), but not all of them, in order to be able to add nodes to an existing cluster. Without at least one other node, the required information cannot be  looked up.

I hope I've understood what you mean correctly?

bootswithdefer:
`{% for host in groups[consul_group_name] %}`
`{% if host != inventory_hostname and (hostvars[host]['consul_node_role'] == 'server' or hostvars[host]['consul_node_role'] == 'bootstrap') %}`
`{{ comma_nodes() }}"{{ hostvars[host][interface]['ipv4']['address'] }}"`

This will require all server/bootstrap hosts in the consul_group_name group to have their facts gathered or hostvars[host][interface] will throw an error.

violuke:
Good point. Possible solutions to this issue would include:
1. Use [Fact Caching](http://docs.ansible.com/ansible/playbooks_variables.html#fact-caching) with Redis or a JSON file.
2. Add to your playbook at the top:
```
- hosts: your_consul_group_name_value
  tasks: [ ]  # No tasks, just gather facts
```
Then change your main play to be more like:
```
- hosts: "{{ limit_hosts|default('your_consul_group_name_value') }}"
  tasks:
    ... # Your actual tasks
```
and run with `ansible-playbook --extra-vars "limit_hosts=myHost" myplaybook.yml`

I think that the code as-is is how it should be. At the end of the day, we'll need to know the ip address of at least one already setup host in the main group and knowing all of them is much better for HA.

bootswithdefer:
Here is another method that works with --limit and doesn't need fact caching:

```---
- hosts: all
  pre_tasks:
    - setup:
      delegate_to: "{{ item }}"
      with_items: "{{ groups['consul-servers'] }}"
      register: consul_server_facts
      run_once: true

    - set_fact: consul_server_addresses="{{ consul_server_facts.results | map(attribute='ansible_facts.ansible_eth0.ipv4.address') | list }}"
      run_once: true

  tasks:
    - debug: var=consul_server_addresses

```

Issue #21, opened 2016-12-20
Role does not work on second run


itewk:
When I try to run the role a second time I get the following error.

**Playbook**:
```
- name: Install & Configure Consul Cluster
  hosts: consuls
  become: yes 
  roles:
    - role: consul
      consul_iface: eth0
      consul_group_name: consuls
```

**Error**:
```
 TASK [consul : Writing key locally to share with other servers that are new] ***
fatal: [consul-host-0]: FAILED! => {"failed": true, "msg": "Failed to get information on remote file (/tmp/consul_raw.key): MODULE FAILURE"}
fatal: [consul-host-2]: FAILED! => {"failed": true, "msg": "Failed to get information on remote file (/tmp/consul_raw.key): MODULE FAILURE"}
fatal: [consul-host-1]: FAILED! => {"failed": true, "msg": "Failed to get information on remote file (/tmp/consul_raw.key): MODULE FAILURE"}
  to retry, use: --limit @/var/lib/awx/projects/_6__consulting_bootcamp/ansible/playbooks/setup_consul.retry
```

I was able to resolve this by setting `become: no` on the local_action tasks.

itewk:
verified merge resolves issue.

PR #22, opened 2016-12-20
set 'become: no' on 'local_action's.

itewk:
Resolves Issue #21


brianshumate:
Hi Ian,

Thanks and this looks good!

itewk:
Thanks for the quick turn around @brianshumate

PR #26, opened 2017-03-07
Initial SmartOS support

sperreault:
I would like to add SmartOS support to your module.


brianshumate:
Fantastic, and thank you!

PR #27, opened 2017-03-10
Updating default consul_bind_address to sane value

arledesma:
This appears to have been broken in 47c9589a5cb16f8a71c504031ba12fb93e135ca3

ipv4 is an object, not an array.
defaults back to address property of ipv4 from consul_iface

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -17,7 +17,7 @@ consul_domain: "consul"
 consul_log_level: "INFO"
 consul_syslog_enable: true
 consul_iface: "eth1"
-consul_bind_address: "{{ hostvars[inventory_hostname]['ansible_'+ consul_iface ]['ipv4'][0].address }}"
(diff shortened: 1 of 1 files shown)


brianshumate:
Ugh, sorry I missed that, thanks!

brianshumate:
Just coming back here to note that ipv4 is actually returned as a list on BSD and SmartOS, and Ansible doesn't handle this properly, see: [Ansible GH-4578](https://github.com/ansible/ansible/issues/4578).

So some kludge in the playbook will be necessary to have this role continue to properly support non Linux OS.

arledesma:
@brianshumate that is interesting, and it has an open bug of ansible/ansible#16615 (initially identified as an issue in ansible/ansible#842).

I think that covering this scenario could get ugly as the ansible team have, for years, not wanted to fully address it because of fears of breaking existing playbooks or duplicating facts.  Best bet is to document it in the README and add additional tests for non-linux variants.

PR #28, opened 2017-03-10
Inconsistent consul_read_key - slurp and deserialize existing bootstrap/config.json

arledesma:
The shell variant was not consistent and error prone with json data.

Slurp, b64decode and deserialize the config from json.


brianshumate:
Very nice, thanks!

PR #29, opened 2017-03-11
Feature support arm

lanefu:
I modified defaults to use a lookup table to translate `ansible_architecture` to the appropriate hashi naming convention so that the role may be used with arm.   The table can easily be added for other `ansible_architecture` returned facts

I've tested successfully tested on Arm7 (OrangePI Plus 2e w/ AllWinner H3) Runnning Armbian Ubuntu Xenial.


brianshumate:
This is great and thank you! I'll review it soon.

PR #30, opened 2017-03-16
Correcting start_join host list in clients

violuke:
I know this changed recently to what is in master currently, but I think the change is wrong and it should be the same as in the `templates/config_server.json.j2` file. My version here seemed to solve my problem locally. Thanks.


brianshumate:
Thanks!

PR #37, opened 2017-03-23
Update installation tags to allow for targeted runs

arledesma:
Some facts and vars are required for installation that were not always available.


brianshumate:
Cool, thank you!

Issue #41, opened 2017-04-04
Consul encryption key is not retrieved from existing json config


misho-kr:
I have 1 server and N clients, and I am running the playbook repeatedly. On subsequent runs the following error is manifested:

```
TASK [brianshumate.consul : Read key for servers that require it] **************
skipping: [10.180.45.22]
 [WARNING]: Unable to find '/tmp/consul_raw.key' in expected paths.

fatal: [10.180.45.18]: FAILED! => {"failed": true, "msg": "could not locate file in lookup: /tmp/consul_raw.key"}
```

It appears the playbook correctly detects an exiting config file on the bootstrapping server (10.180.45.18), but it does not read it from that server. Instead it tries to read it from host that is not designated as bootstrapping server (10.180.45.22). 

I suspect the culprit is the playbook uses _run_once_ clause in a block where the block has _when_ condition. What happens is _run_once_ clause is applied first, it picks one host which may or may not be a bootstrapping server, and then the _when_ clause is applied. If that host is not a server then there is no config to read and later on the playbook fails with the above message.

I was able to get pass this error by removing the _run_once_ clause from the tasks enclosed in block. I will create a PR shortly for review.

brianshumate:
Thanks for the PR! I'll go ahead and close this now, too.

PR #42, opened 2017-04-04
look for existing consul config on all hosts

misho-kr:
Fix for #41 

Task to read consul config file should be executed on all hosts, otherwise _run_once_ may select to run the task on host that is not a server. This may happen because apparently Ansible evaluates _run_once_ first and then it checks the _when_ clause.


brianshumate:
Thanks!

PR #47, opened 2017-04-09
Fix wrong when statement preventing key transfer

### LINKED ISSUE #46: Key not read by servers that need it
Hi Brian,

Thanks for your work on this role!

When adding a server after initial setup, the key is not correctly read by the server.

I think it's a wrong check;
https://github.com/brianshumate/ansible-consul/blob/master/tasks/main.yml#L125

```yaml
- name: Read key for servers that require it
  set_fact:
    consul_raw_key: "{{ lookup('file', '/tmp/consul_raw.key') }}"
  when: consul_raw_key is not defined and bootstrap_marker.stat.exists
```
Servers that require it don't have the `bootstrap_marker` file. So I think the check should be `... and not bootstrap_marker.stat.exists`.

Happy to provide a PR if you can confirm.

groggemans:
Also introduces list notation for longer when statements. This makes the
statements more clear and prevents future logic errors.

Fixes #46

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -9,22 +9,30 @@
 - name: Fail if not a new release of CentOS or Red Hat
   fail:
     msg: "{{ ansible_distribution_version }} isn't a supported version."
-  when: ansible_distribution in ['RedHat', 'CentOS'] and ansible_distribution_version|version_compare(6, '<')
+  when:
+    - ansible_distribution in ['RedHat', 'CentOS']
+    - ansible_distribution_version|version_compare(6, '<')
 
 - name: Fail if not a new release of Debian
   fail:
     msg: "{{ ansible_distribution_version }} is not a supported version."
-  when: ansible_distribution == "Debian" and ansible_distribution_version|version_compare(8.5, '<')
+  when:
+    - ansible_distribution == "Debian"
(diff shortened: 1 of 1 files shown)


tasks/main.yml
```diff
@@ -112,22 +126,31 @@
...
+    - consul_raw_key is defined
+    - bootstrap_marker.stat.exists
 
 - name: Read key for servers that require it
   set_fact:
     consul_raw_key: "{{ lookup('file', '/tmp/consul_raw.key') }}"
-  when: consul_raw_key is not defined and bootstrap_marker.stat.exists
+  when:
+    - consul_raw_key is not defined
+    - not bootstrap_marker.stat.exists
```
groggemans, line 146: Added the `not` here to fix the bug.

Should have done it in a separate commit, but thought too late about it.

groggemans:
@brianshumate I'm running into an issue with further testing. Best to wait with merging until I figure it out.

brianshumate:
Thanks again!

groggemans:
@brianshumate Looks like you missed the comment.

It fixed setting up a server after initial setup, but broke the initial setup. Looks like a bigger logic change is necessary here to fix the issue for all possible situations.

I'm looking into it. You may wan't to revert the PR in the mean time. I hope to have a fix later this evening.

PR #49, opened 2017-04-09
Fix bootstrapping

### LINKED ISSUE #46: Key not read by servers that need it

groggemans:
If a key is provided through the extra vars or inventory, this is used. If no key was provided, the server/bootstrap nodes will be checked. If still no key was written to disk, a new key will be generated. All nodes will then use the key written to disk. This approach also allows a key change on all nodes.

The key is always temporary written to disk. Still not perfect, but it should do the job for now.

Tested on a one and a three node cluster. Also tested replacing one of the servers in the three node cluster.

Did some cleanup to, hopefully more in a future PR.

Sorry  for the trouble with the previous PR, this one hopefully doesn't introduce a bigger bug then it fixes!

Fixes #46 
Replaces #47

### DIFF
--- a/tasks/asserts.yml
+++ b/tasks/asserts.yml
@@ -1,0 +1,62 @@
+---
+# File: checks.yml - Asserts for this playbook
+
+- name: Check distribution compatibility
+  fail:
+    msg: "{{ ansible_distribution }} is not currently supported by this role."
+  when: ansible_distribution not in ['RedHat', 'CentOS', 'Debian', 'FreeBSD', 'SmartOS', 'Ubuntu']
+
+- name: Fail if not a new release of CentOS or Red Hat
+  fail:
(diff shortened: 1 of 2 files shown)


brianshumate:
👍 Great!

PR #51, opened 2017-04-11
adding flag and logic to install binaries on remote hosts instead of …

jonhatalla:
…on the control machine


brianshumate:
Looks good, thanks!

Issue #40, opened 2017-03-31
Issues with consul_iface


AlJaMa:
Hello! 

Thank you for the module! One thing I am not understanding is the "consul_iface" variable. 
Several of my machines do not have the same interface name, so I have attempted to set "consul_iface" to "{{ ansible_default_ipv4.interface }}"  

When your module runs, it then fails with the following error.

```
fatal: [10.11.1.46]: FAILED! => {
    "changed": false, 
    "failed": true, 
    "invocation": {
        "module_args": {
            "dest": "/etc/consul.d/client/config.json", 
            "group": "bin", 
            "owner": "consul", 
            "src": "config_client.json.j2"
        }, 
        "module_name": "template"
    }, 
    "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute u'ansible_officelanbr0'"
}
```
It appears that https://github.com/brianshumate/ansible-consul/blob/master/templates/config_client.json.j2#L29 is properly trying to find the var (see below for the output of `ansible -m setup`

```
       "ansible_officelanbr0": {                                                                                                                           
            "active": true,                                                                                                                                 
            "device": "officelanbr0",                                                                                                                       
            "features": {                                                                                                                                   
                "busy_poll": "off [fixed]",                                                                                                                 
                "fcoe_mtu": "off [fixed]",                                                                                                                  
                "generic_receive_offload": "on",                                                                                                            
                "generic_segmentation_offload": "on",                                                                                                       
                "highdma": "on",                                                                                                                            
                "hw_tc_offload": "off [fixed]",                                                                                                             
                "l2_fwd_offload": "off [fixed]",                                                                                                            
                "large_receive_offload": "off [fixed]",                                                                                                     
                "loopback": "off [fixed]",                                                                                                                  
                "netns_local": "on [fixed]",                                                                                                                
                "ntuple_filters": "off [fixed]",                                                                                                            
                "receive_hashing": "off [fixed]",                                                                                                           
                "rx_all": "off [fixed]",                                                                                                                    
                "rx_checksumming": "off [fixed]",                                                                                                           
                "rx_fcs": "off [fixed]",                                                                                                                    
                "rx_vlan_filter": "off [fixed]",                                                                                                            
                "rx_vlan_offload": "off [fixed]",                                                                                                           
                "rx_vlan_stag_filter": "off [fixed]",                                                                                                       
                "rx_vlan_stag_hw_parse": "off [fixed]",                                                                                                     
[...]

PR #52, opened 2017-04-16
Cleanup and improvements

groggemans:
Lot's of cleanup, tweaks and improvements in this PR.

Did a cleanup of the templates to produce cleaner output. Also added the `consul_servers` variable to make it easier to loop over them in the templates.

Did some cleanup in the defaults file, and changed to default value of `consul_iface` to always contain a valid interface by default (see comment in #40 )

Added the 'gather server facts' task to discover the `consul_bind_address` from servers that are currently not targeted (not used by default). Also fixed the `consul_bind_address` default value so it works on FreeBSD again.

It's now possible to bootstrap a client or server by only targeting that node if the 'gather server facts' task is enabled or all servers have there `consul_bind_address` set in there hostvars file/the inventory. (And the gossip key is provided in host/group vars or env.)

Added *client* as default for the `consul_node_role` as it makes sense to bootstrap a client by default.

Added a switch (default *True*) for the user management task and placed it in it's own file. Also hinted on more configurable group management for the future.

Moved the `vault_address` var to the consul namespace (`consul_vault_address`) with a default value of `vault_address | default ('0.0.0.0', true)`. This preserves the current behavior but makes it possible to have a different vault address for consul then the default vault setup for the node.

Moved directory settings to there own file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ inventory file (see below):
 
 - Consul network interface
   - Override with `CONSUL_IFACE` environment variable
-- Default value: determined from host inventory
+- Default value: `{{ ansible_default_ipv4.interface }}`
 
 ### `consul_bind_address`
 
@@ -301,6 +301,25 @@ inventory file (see below):
 - Whether to download the files for installation directly on the remote hosts
 - Default value: *false*
 
+### `consul_node_role`
+
+- Consul server role, one of: *bootstrap*, *server*, or *client*
+- Default value: *client*
+
+One server should be designated as the bootstrap
+server that uses the `bootstrap_expect` configuration directive, and the other
+servers will connect to this server. You can also specify *client* as the
+role, and Consul will be configured as a client agent instead of a server.
+
+Here is an example of how the hosts inventory could be defined for a simple
+cluster of 3 servers:
+
+```
+consul1.local consul_node_role=bootstrap
+consul2.local consul_node_role=server
+consul3.local consul_node_role=server
+```
+
 #### Custom Configuration Section
 
 As Consul loads the configuration from files and directories in lexical order,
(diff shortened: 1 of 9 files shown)


brianshumate:
Thanks for all this. I'll go ahead and merge, but might need to change some things.

I am not sure if I want to leave client as the default node role, for example. The role is primarily for servers and clustering them.

groggemans:
> I am not sure if I want to leave client as the default node role, for example. The role is primarily for servers and clustering them.

That sounds a little strange to me. The role manages consul in general, so why would we only focus on servers? The people using this role probably also need clients to be bootstrapped/managed (I do).

As you only need a few servers, and probably attach a lot of clients, defaulting to client seems logical.

brianshumate:
Yes, when I originally created the role, the emphasis was on quick clusters of servers for testing and so on and for this it does well. The more generalized use case which does require more clients than servers in almost every case is probably best if this is to be a more broadly useful role.

I can just update the `Vagrantfile` to pass the right var to make servers in that case, too.

Definitely thanks for all the work in this PR!

brianshumate:
It looks like by moving `consul_node_role` into defaults it cannot be specified in host inventory any longer because defaults override host vars and now the role can no longer bootstrap a cluster because of the default `consul_node_role` being client.

The  `consul_node_role` var needs to have a _per host_ value since one host must be designated as the bootstrap server. The var cannot default to just one value for this reason.

I think changing this to a default value pretty much breaks my primary design intent for this role which is actually _not to maintain long living clusters_ (that's a secondary benefit really) but to bootstrap new clusters of servers.

I am going to have to revert the `consul_node_role change` in light of this problem and make sure the primary source comes from the host inventory.

groggemans:
That doesn't sound correct, see the ansible variable precedence;

```
In 2.x, we have made the order of precedence more specific (with the last listed variables winning prioritization):

role defaults [1]
inventory INI or script group vars [2]
inventory group_vars/all
playbook group_vars/all
inventory group_vars/*
playbook group_vars/*
inventory INI or script host vars [2]
inventory host_vars/*
playbook host_vars/*
host facts
play vars
play vars_prompt
play vars_files
role vars (defined in role/vars/main.yml)
block vars (only for tasks in block)
task vars (only for the task)
role (and include_role) params
include params
include_vars
set_facts / registered vars
extra vars (always win precedence)
```
Role defaults are overruled by **all** other variables!

Are you sure nothing else is messing with the variable?

brianshumate:
Ahh yes, I misread the docs. It looks like host vars do override the defaults after all. 😄 

Sorry for needlessly reverting your work. I'll put it back in there.

groggemans:
NP Brian! :)

I also couldn't seem to reproduce your problem on my test cluster. Are you using the included vagrant setup for testing?

PR #56, opened 2017-04-18
Config dirs files

### LINKED ISSUE #53: Config files and dir structure
Currently the following config dir structure is used;
```
/etc/consul
└── .consul_bootstrapped
/etc/consul.d
├── bootstrap
│   └── config.json
├── client
│   └── config.json
└── server
    └── config.json
```
I would like to change this to a more common layout where the agent config resides in the `consul` dir. And all files in `consul.d` are included automatically.
```
/etc/consul
├── {{ consul_node_role }}.json
└── .consul_bootstrapped
/etc/consul.d
└── random_config.json
```
In this example I only added the config for the specific role of the agent. Is there a reason why the role currently provides all 3 configs? I would provide bootstrap/server to all servers and only client to clients.

Happy to provide a PR.

### LINKED ISSUE #55: consul_servers only works if consul_node_role is a fact
I think I found the problem you were having with the `consul_node_role`. When a default is used, the variable isn't available as a fact. But the used loops and look-ups need this as a fact.

Simply setting it as a fact should solve the problem.

I'll add a fix to the config layout PR

groggemans:
Added `consul_debug` variable to toggle debugging/testing options. Defaults to `no`.

Updated `consul_config_path` to point to */etc/consul* and added `consul_configd_path` defaulting to */etc/consul.d*.

Moved all config related tasks to the `tasks/config.yml` file.

Added the ACL and TLS parameters to the main `config.json` file that will be created. Using `/etc/consul/config.json` for all consul roles makes a few things easier. It also fixes a small bug that prevented the gossip key to be read from normal servers (previously bootstrap only).

If the `consul_debug` variable is set to yes/true, all config files will be generated in the config dir (eg. `/etc/consul/config.json`, `.../bootstrap.json`, `.../server.json`, `.../client.json`)

The custom config and acl rules will be added to the `config.d` dir and are prefixed with the number 50. So it uses the common numbering convention to change load order.

Updated all init files to use the new config file/dir layout. Best to have a good look as I'm not 100% familiar with all the different init systems and there files/syntax.

Exposed `consul_node_role` as a fact. ( Fixes #55 )

Closes #53


brianshumate:
👍 It looks good, thanks!

PR #61, opened 2017-04-27
Add bootstrap-expect toggle option

### LINKED ISSUE #54: bootstrap without bootstrap node
If I read the info about the `-bootstrap-expect` option in the consul docs, I'm under the impression this is the better choice for normal clusters (not for testing that is).

As this doesn't require a node to be in bootstrap mode all the time, or restart it in normal server mode after bootstrapping. (something that's currently not done by this role, but would be good practice if I understand correctly)

I'd like to add an option to choose between both bootstrapping techniques.

What's your opinion on this @brianshumate?

groggemans:
Closes #54 

Adds an option to bootstrap using the `bootstrap_expect` option instead of a bootstrap server (using the `bootstrap` option).

Defaults to using a bootstrap server for now, we could still change this.

Also adds some internal variables (`_consul_bootstrap_servers` and `_consul_bootstrap_server`) for internal usage. Currently only used for two new asserts that check if there's max 1 bootstrap server and no bootstrap servers are defined while the `bootstrap_expect` option is used.


brianshumate:
Thanks!

PR #62, opened 2017-04-28
Generate correct JSON with TLS and ACL enabled

tbartelmess:
This is more or less a dirty hack, maybe Jinja has a better way to append json and conditionally add the comma after `"ui": true`.


groggemans:
Oops, looked over those comma's when I changed the template.

Fix looks good.

PR #66, opened 2017-05-07
Multi datacenter support

groggemans:
Adds some extra internal variables to differentiate between lan servers and wan servers. This allows for basic multi-datacenter support.

Hosts in the same group will use the same encryption key. Defining multiple datacenters in the same group results in multiple auto joining datacenters.

### DIFF
--- a/templates/config_server.json.j2
+++ b/templates/config_server.json.j2
@@ -17,7 +17,7 @@
   {% endif -%}
   "bootstrap": false,
   {% if consul_bootstrap_expect -%}
-    "bootstrap_expect": {{ consul_servers|length }},
+    "bootstrap_expect": {{ _consul_lan_servers|length }},
   {% endif -%}
   "server": true,
   "node_name": "{{ consul_node_name }}",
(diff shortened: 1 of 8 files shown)


groggemans:
I've tested this with bootstrap-expect, but I forgot something for bootstrapping with the default bootstrap option. (It still checks for max 1 bootstrap server over all servers) I'll make some modifications to fix this.

groggemans:
After that last commit it should work for both bootstrap options.

templates/config_server.json.j2
```diff
@@ -17,7 +17,7 @@
   {% endif -%}
   "bootstrap": false,
   {% if consul_bootstrap_expect -%}
-    "bootstrap_expect": {{ consul_servers|length }},
+    "bootstrap_expect": {{ _consul_lan_servers|length }},
```
brianshumate: Minor thing, but can you add a spaces on either side of the pipe please?

brianshumate (changes requested):
Awesome, this looks great after the one little style fix.

Thanks!

templates/config_server.json.j2
groggemans: @brianshumate do you want to have spaces in these cases to?

`{% if _consul_wan_servers|length > 0 -%}`

Or only when the result is not used in an expression?

groggemans:
I've added the spaces around all pipe symbols/filters.

Also updated the bootstrap and client templates, seems I forgot those. We may want to try to combine them in a single template file to prevent this in the future.

brianshumate:
👍 Excellent! Thanks again. Gertjan!

PR #68, opened 2017-05-10
update (set ownership of) consul folders recursively

misho-kr:
This playbook creates (or updates existing) several folders for Consul and sets ownership to consul user and group.

Files inside those folders  (for example keyring files) should be updated too by executing the Ansible task with "recursive" option.


brianshumate:
👍 Thanks!

PR #69, opened 2017-05-17
Windows support!

judy-zz:
I wrote full Windows support for this role. I tried to match the programming style and Ansible best practices as closely as possible. I'm still new to Ansible, and would happily accept any feedback.

There are a LOT of additions, and more than a handful of changes, mostly `ansible_os_system` checks. I'll comment in the code diff where there might be points of confusion between how Linux does things and how Windows does them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,7 @@ specific software and versions:
 * FreeBSD: 11
 * RHEL: 7
 * Ubuntu: 16.04
-
-Sorry, there is no planned support at the moment for Windows.
+* Windows: Server 2012 R2
 
 ## Role Variables
 
@@ -86,37 +85,44 @@ the variables are named and described below:
 ### `consul_bin_path`
 
 - Binary installation path
-- Default value: `/usr/local/bin`
+- Default linux value: `/usr/local/bin`
+- Default windows value: `C:\ProgramData\consul\bin`
 
 ### `consul_config_path`
 
 - Base configuration file path
-- Default value: `/etc/consul`
+- Default linux value: `/etc/consul`
+- Default windows value: `C:\ProgramData\consul\config`
 
 ### `consul_configd_path`
 
 - Additional configuration directory
-- Default value: `/etc/consul.d`
+- Default linux value: `/etc/consul.d`
+- Default windows value: `C:\ProgramData\consul\config.d`
 
 ### `consul_data_path`
 
 - Data path
-- Default value: `/var/consul`
+- Default linux value: `/var/consul`
+- Default windows value: `C:\ProgramData\consul\data`
 
 ### `consul_log_path`
 
 - Log path
-- Default value: `/var/log/consul`
+- Default linux value: `/var/log/consul`
+- Default windows value: `C:\ProgramData\consul\log`
 
 ### `consul_run_path`
 
 - Run path for PID file
-- Default value: `/var/run/consul`
+- Default linux value: `/var/run/consul`
+- Default windows value: `C:\ProgramData\consul`
 
 ### `consul_user`
 
 - OS user
-- Default value: *consul*
+- Default linux value: *consul*
+- Default linux value: *LocalSystem*
 
 ### `consul_group`
 
@@ -170,7 +176,8 @@ in many Ansible versions, so this feature might not always work.
 
 - Log to syslog
   - Override with `CONSUL_SYSLOG_ENABLE` environment variable
-- Default value: *true*
+- Default linux value: *true*
+- Default windows value: *false*
 
 ### `consul_iface`
 
@@ -347,6 +354,7 @@ in many Ansible versions, so this feature might not always work.
 ### `consul_install_remotely`
 
 - Whether to download the files for installation directly on the remote hosts
+- This is the only option on Windows as WinRM is somewhat limited in this scope.
 - Default value: *false*
 
 ### `consul_node_role`
@@ -474,6 +482,26 @@ packages with different package names.
 - List of OS packages to install
 - Default value: list
 
+### `consul_windows_pkg`
+
+- Consul package filename
+- Default value: `{{ consul_version }}_windows_amd64.zip`
+
+### `consul_windows_url`
+
+- Consul package download URL
+- Default value: `{{ consul_zip_url }}`
+
+### `consul_windows_sha256`
+
+- Consul download SHA256 summary
+- Default value: SHA256 SUM
+
+### `consul_windows_os_packages`
+
+- List of OS packages to install
+- Default value: list
+
 ## Dependencies
 
 Ansible requires GNU tar and this role performs some local use of the
@@ -482,6 +510,11 @@ in the PATH.
 
 If you're on system with a different (i.e. BSD) `tar`, like macOS and you
 see odd errors during unarchive tasks, you could be missing `gtar`.
+
+Installing Ansible on Windows requires the PowerShell Community Extensions.
+These already installed on Windows Server 2012 R2 and onward. If you're
+attempting this role on Windows Server 2008 or earlier, you'll want to install
+the extensions [here](https://pscx.codeplex.com/).
 
 ## Example Playbook
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -17,7 +17,7 @@ consul_architecture_map:
   64-bit: amd64
 consul_architecture: "{{ consul_architecture_map[ansible_architecture] }}"
 consul_os: "\
-  {% if ansible_os_family == "Windows" %}\
+  {% if ansible_os_family == 'Windows' %}\
     {{ 'windows' }}\
   {% else %}\
     {{ ansible_system | lower }}\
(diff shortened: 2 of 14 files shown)


README.md
```diff
@@ -86,37 +85,44 @@ the variables are named and described below:
 ### `consul_bin_path`
 
 - Binary installation path
-- Default value: `/usr/local/bin`
+- Default linux value: `/usr/local/bin`
+- Default windows value: `C:\ProgramData\consul\bin`
```
judy-zz, line 89: I've used `C:\ProgramData\consul` for all necessary program data, including the consul.exe binary. Most Windows programs use the magical `%APPDATA%` path, but those are tied to specific users. My understanding is that `C:\ProgramData` is what Chocolatey and NSSM use, and it's a good place for system-level services.

README.md
```diff
@@ -86,37 +85,44 @@ the variables are named and described below:
...
-- Default value: `/var/run/consul`
+- Default linux value: `/var/run/consul`
+- Default windows value: `C:\ProgramData\consul`
 
 ### `consul_user`
 
 - OS user
-- Default value: *consul*
+- Default linux value: *consul*
+- Default linux value: *LocalSystem*
```
judy-zz, line 125: Windows convention is to have the `LocalSystem` user run services. It has the right mix of permissions for those kinds of services.

defaults/main.yml
```diff
@@ -5,20 +5,26 @@
 consul_debug: no
 
 ### Package
-consul_version: "{{ lookup('env','CONSUL_VERSION') | default('0.8.2', true) }}"
+consul_version: "{{ lookup('env','CONSUL_VERSION') | default('0.8.3', true) }}"
```
judy-zz, line 8: The last commit on the upstream repo updates the project to support Consul 0.8.3, but the default installation still downloads 0.8.2. This is an unrelated bug fix, but will probably just disappear if fixed upstream, correct?

defaults/main.yml
```diff
@@ -5,20 +5,26 @@
...
 consul_architecture_map:
   # this first entry seems... redundant
   #(but it's required for reasons)
   amd64: amd64
   x86_64: amd64
   armv7l: arm
   aarch64: arm64
+  32-bit: "386"
```
judy-zz, line 16: I don't know if there are even Windows Server versions that run in 32-bit, but there are personal versions of Windows that do, and this entry ensures the right package is grabbed from Hashicorp.

defaults/main.yml
```diff
@@ -5,20 +5,26 @@
...
+  64-bit: amd64
 consul_architecture: "{{ consul_architecture_map[ansible_architecture] }}"
-consul_os: "{{ ansible_system | lower }}"
+consul_os: "\
+  {% if ansible_os_family == "Windows" %}\
+    {{ 'windows' }}\
+  {% else %}\
+    {{ ansible_system | lower }}\
```
judy-zz: `ansible_system` in these cases returns **Win32NT**. This changes between Windows Server versions and is unfortunately brittle.

defaults/main.yml
```diff
@@ -27,6 +33,7 @@ consul_install_remotely: false
 consul_bin_path: "/usr/local/bin"
 consul_config_path: "/etc/consul"
 consul_configd_path: "/etc/consul.d"
+consul_bootstrap_state: "/etc/consul/.consul_bootstrapped"
```
judy-zz, line 36: Bootstrap state is another path I needed to change for Windows, and consolidating it with the other paths seemed to be the smart choice.

defaults/main.yml
```diff
@@ -43,7 +50,12 @@ consul_datacenter: "{{ lookup('env','CONSUL_DATACENTER') | default('dc1', true)
 consul_domain: "{{ lookup('env','CONSUL_DOMAIN') | default('consul', true) }}"
 consul_log_level: "{{ lookup('env','CONSUL_LOG_LEVEL') | default('INFO', true) }}"
 consul_syslog_enable: "{{ lookup('env','CONSUL_SYSLOG_ENABLE') | default(true, true) }}"
-consul_iface: "{{ lookup('env','CONSUL_IFACE') | default(ansible_default_ipv4.interface, true) }}"
+consul_iface: "\
+  {% if ansible_os_family == "Windows" %}\
+    {{ lookup('env','CONSUL_IFACE') | default(ansible_interfaces[0].interface_name, true) }}\
```
judy-zz: Network interfaces in Windows are handled very differently in Ansible, with very little overlap. I'm not sure why. But this works.

handlers/main.yml
```diff
@@ -2,15 +2,10 @@
 # File: main.yml - Handlers for Consul
 
 - name: restart consul
-  service:
-    name: consul
-    state: restarted
+  include: restart_consul.yml
```
judy-zz, line 5: Having OS-contextual handlers is awkward. I attempted to do this with ansible blocks but they're unsupported. This is clunkier than ideal, but is the best option I could find that works.

tasks/install_windows.yml
```diff
@@ -0,0 +1,63 @@
+---
+# File: install_remote.yml - package installation tasks for Consul
+
+- name: Validate remote Consul directory
+  win_file:
+    path: /tmp/consul
+    state: directory
+
+- name: Verify TLS1.2 is used
```
judy-zz, line 9: Windows attempts to use TLS1.0 by default when downloading files, falling back to SSL3, and never trying TLS1.2. This is fixed in later versions of Windows.

This task sets a typical value in the registry just in case it isn't there, ensuring that TLS1.2 is tried first, *then* TLS1.0, *then* SSL3. Otherwise, downloading the package and checksum files outright fails because you haven't changed the registry variable. ¯\_(ツ)_/¯

tasks/install_windows.yml
```diff
@@ -0,0 +1,63 @@
...
+
+- name: Download Consul package checksum file
+  win_get_url:
+    url: "{{ consul_checksum_file_url }}"
+    dest: "/tmp/consul/consul_{{ consul_version }}_SHA256SUMS"
+  tags: installation
+  when: not consul_checksum.stat.exists | bool
+
+- name: Read Consul package checksum
+  win_shell: "findstr {{ consul_pkg }} /tmp/consul/consul_{{ consul_version }}_SHA256SUMS"
```
judy-zz, line 30: `findstr` is to Windows as `grep` is to Linux.

tasks/install_windows.yml
```diff
@@ -0,0 +1,63 @@
...
+  register: consul_pkg_checksum
+  tags: installation
+
+- name: Download Consul
+  win_get_url:
+    url: "{{ consul_zip_url }}"
+    dest: "/tmp/consul/{{ consul_pkg }}"
+  tags: installation
+
+- name: Calculate checksum
```
judy-zz, line 40: These next two tasks are done different from the usual path, because `win_get_url` doesn't allow checksums like `get_url` does. Minor inconsistency we have to handle by hand here.

tasks/main.yml
```diff
@@ -129,6 +160,18 @@
 
 - name: Create Consul configuration
   include: config.yml
+  when: ansible_os_family != 'Windows'
+
+- name: Create Consul configuration for Windows
+  include: config_windows.yml
+  when: ansible_os_family == 'Windows'
+
+- name: Ensure neither ACL nor TLS are requested for windows hosts
```
judy-zz, line 169: Small safety check. I haven't updated the ACL or TLS code, so in the meanwhile their support is turned off for Windows hosts. This can be added later.

tasks/main.yml
```diff
@@ -157,10 +200,12 @@
       group: root
       mode: 0755
     when:
+      - ansible_service_mgr is defined
```
judy-zz, line 203: `ansible_service_mgr` is undefined on Windows.

tasks/main.yml
```diff
@@ -215,16 +264,48 @@
...
       enabled: yes
+    when: ansible_os_family != "Windows"
+
+  - name: Install NSSM
+    win_chocolatey:
+      name: nssm
+      state: latest
+
+  - name: Start Consul on Windows
+    win_nssm:
```
judy-zz, line 275: NSSM is used to wrap consul.exe into a Windows-friendly service. Without it we can't use the regular `win_service` module because services don't let you set configuration variables.

tasks/main.yml
```diff
@@ -215,16 +264,48 @@
...
+      stderr_file: "{{ consul_log_path }}/consul-nssm-error.log"
+    when: ansible_os_family == "Windows"
 
   - name: Check Consul HTTP API
     wait_for:
       delay: 15
       port: 8500
+    when: ansible_os_family != "Windows"
+
+  - name: Check Consul HTTP API
```
judy-zz, line 290: There's no `win_wait_for` module yet. This is a hack to use `wait_for` locally and watch the port in the same way. It works on my machine perfectly, but I can see cases where this could potentially fail depending on network setup. It may need more work in the future.

vars/Windows.yml
```diff
@@ -0,0 +1,17 @@
+---
+# File: Windows.yml - Windows OS variables for Consul
+
+# paths
+consul_windows_path: /ProgramData/consul
```
judy-zz, line 5: Fun fact: forward slashes now work almost as well as backslashes do in Windows! Folder structure is still different though.

brianshumate:
Thanks very much for this contribution!

tasks/main.yml
```diff
@@ -215,16 +264,48 @@
       name: consul
       state: started
       enabled: yes
+    when: ansible_os_family != "Windows"
+
+  - name: Install NSSM
+    win_chocolatey:
+      name: nssm
+      state: latest
```
groggemans, line 272: @judy @brianshumate Shouldn't there be a when clause here checking for os_family == "Windows"?
(Didn't test)

gmarkey:
Heads up; the "when" clauses are breaking anything that registers objects in the output (e.g. stat) as the condition failure is being stored in register and forcibly casting as a dict.

Might just be an ordering problem; it's likely best to encapsulate the Windows & non-Windows stuff inside a block.

brianshumate:
Yeah... missed that stuff, sorry!

I have committed a quick change that puts all non-Windows and Windows main tasks in separate top-level blocks. Not the ideal solution, but it unbreaks the role for the moment. This is released in v1.20.1 also.

judy-zz:
I was trying this earlier but wasn't formatting my `blocks` correctly. I see how you did it now. This is much cleaner!

brianshumate:
👍

PR #70, opened 2017-06-03
Combine config templates

groggemans:
Combines all (client/server/bootstrap) config templates so we don't end up with differing configs between node roles. (eg missing parameters or different defaults)

I also switched the template module for the copy module and a template lookup. This allows us to pass the template content through the `to_nice_json` filter. As a result we don't need to think about the formatting or trailing comma's when editing the template and it's contents.

By adding an additional `with_items` variable I removed some duplicate code needed for the debug feature that adds all config types.

I also added the `consul_ui` parameter to optionally disable the ui.


brianshumate:
This is great, thank you very much!

PR #73, opened 2017-06-22
improve HTTP API check

dmke:
The check whether Consul had started failed when `consul_ports.http` was set to something other than 8500. It also failed, if you set the `consul_http_bind_address` to a Unix socket path.

This improves the check in *NIX block in `tasks/main.yml` by wrapping it in a conditional:

- if `consul_ports.http` > 0, *and* `consul_http_bind_address` is a valid IP address, use `wait_for`+`port`
- if `consul_http_bind_address` starts with `unix:/`, use `wait_for`+`path`

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -232,10 +232,17 @@
           state: started
           enabled: yes
 
-      - name: Check Consul HTTP API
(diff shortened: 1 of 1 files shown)


brianshumate:
Nice, thanks a lot for this!

PR #78, opened 2017-06-26
Fix runtime warnings

vincent-legoll:
Those warnings were observed with :

    # consul --version
    Consul v0.8.4

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,11 +206,6 @@ in many Ansible versions, so this feature might not always work.
 - HTTPS API bind address
 - Default value: *0.0.0.0*
 
-### `consul_rpc_bind_address`
-
-- RPC bind address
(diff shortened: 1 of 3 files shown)


brianshumate:
This one is going to be tricky as we need to aim for backwards compatibility / the ability to run older Consul versions.

vincent-legoll:
Ouch, care to explain a bit what's required ?
sprinkling something alike: "when: consul_version >= '0.8.x'"

oliverprater:
I have worked out another "fix" am working on putting it into my branch and creating the PR at the moment.

oliverprater:
#86

Here is my proposed solution

vincent-legoll:
OK, your solution will handle both cases, so I'm retracting this one

PR #86, opened 2017-06-28
Fix breaking change in paths and runtime warning

oliverprater:
This fixes the faulty commit 23c6f48 and the runtime warnings in consul version >=0.8 due to rpc configuration issues.

In case my suggested fix for the runtime warnings is not well-rounded enough I can open a separate PR with commit 19bfebf

As the incorrect escaping of the string lead to total failure of the consul agent due to:

> Error starting agent: Failed to start Consul server: Error loading CA File: open {{ consul_config_path }}/ssl/ca.pem: no such file or directory


README.md
```diff
@@ -213,16 +213,34 @@ in many Ansible versions, so this feature might not always work.
 ### `consul_rpc_bind_address`
 
 - RPC bind address
-- Default value: *0.0.0.0*
+- Default value:
+  - consul version  <0.8: *0.0.0.0*
+  - consul verison >=0.8: ""
```
vincent-legoll: typo "verison" + space around <

README.md
```diff
@@ -213,16 +213,34 @@ in many Ansible versions, so this feature might not always work.
...
 ### `consul_ports`
 
 - Port Mappings
-- Default value:
-  - `rpc`: `8400`
-  - `http`: `8500`
-  - `https`: `-1`
-  - `dns`: `8600`
+- The default values are set according to the consul version inside the playbook.
+  - consul verison <0.8:
```
vincent-legoll: idem + space around <

README.md
```diff
@@ -213,16 +213,34 @@ in many Ansible versions, so this feature might not always work.
...
+    - `dns`: `8600`
+    - `http`: `8500`
+    - `https`: `-1`
+    - `rpc`: `8400`
+  - consul verison >=0.8:
```
vincent-legoll: +1

tasks/main.yml
```diff
@@ -15,6 +15,27 @@
...
+    consul_ports:
+      dns: 8600
+      http: 8500
+      https: -1
+      rpc: 8400
+  when:
+    - consul_ports | length == 0
+    - consul_version | version_compare('0.8', '<')
+
+- name: Set default port configuration for the consul version >=0.8
```
vincent-legoll: missing space around ">="

tasks/main.yml
```diff
@@ -15,6 +15,27 @@
 - name: Include checks/asserts
   include: asserts.yml
 
+- name: Set default port configuration for the consul version <0.8
```
vincent-legoll: missing space around "<"

vincent-legoll (commented):
I'm OK with that fix if the small glitches are fixed

oliverprater:
Fixed the typo's in this PR

oliverprater:
Squashed all the typo fixes ... lgtm

groggemans:
I'm not a fan of settings defaults in the tasks. As this makes it hard to keep an overview of all default variables and there values.

We're also adding a lot of code just for ignoring the rpc port value on versions lower then 0.8. I think it would be better to alter the template in this case. This would also allow us to use the same settings for hosts with mixed versions.

Same could be done for the rpc address.

(Please correct me if I've mist something, only took a quick look at the problem/PR )

defaults/main.yml
```diff
@@ -109,7 +114,7 @@ consul_acl_policy: false
 ## TLS
 consul_tls_enable: "{{ lookup('env','CONSUL_TLS_ENABLE') | default(false, true) }}"
 consul_tls_src_files: "{{ lookup('env','CONSUL_TLS_SRC_FILES') | default(role_path+'/files', true) }}"
-consul_tls_dir: "{{ lookup('env','CONSUL_TLS_DIR') | default('{{ consul_config_path }}/ssl', true) }}"
+consul_tls_dir: "{{ lookup('env','CONSUL_TLS_DIR') | default(consul_config_path+'/ssl', true) }}"
```
vincent-legoll: Put spaces around that + operator, please

oliverprater:
I think @groggemans is correct with his observation.

The `defaults/main.yml` seems to be doing the job of the template ... it should be the other way around!

templates/config.json.j2
```diff
@@ -24,8 +24,28 @@
...
     },
-    {% if consul_ports %}
-        "ports": {{ consul_ports | to_json }},
+    {## Ports Used ##}
+    {% if consul_ports | length > 0 %}
+      "ports": {{ consul_ports | to_json }},
+    {% else %}
+      "ports": {
+        {% if _consul_default_port_rpc | trim != '' %}
+          {% set _ = consul_ports.__setitem__('rpc', _consul_default_port_rpc) %}
```
vincent-legoll, line 33: Why the __setitem__() use ? I'd think this needs at least a comment.

oliverprater: The template uses some jinja2 magic also present in the DebOps project to ensure that the consul_ports dict object is always identical to the config.json file values.

See https://stackoverflow.com/questions/36886650/how-to-add-a-new-entry-into-a-dictionary-object-while-using-jinja2
See https://stackoverflow.com/questions/4870346/can-a-jinja-variables-scope-extend-beyond-in-an-inner-block

Just noticed from my comment it is unclear for what reason we need this jinja2 magic.

When you look, for instance, at `tasks/main.yml` line 238 the variable `{{ consul_ports.http|int }}` is used to `wait_for` the HTTP API to become available. The jinja2 magic ensures that this happens as expected.

oliverprater:
In case we plan to continue with this strain of logic, then I can also rework addresses as well. This would introduce the variable `consul_addresses: {}` and deprecate the variables:

```
consul_dns_bind_address: "127.0.0.1"
consul_http_bind_address: "0.0.0.0"
consul_https_bind_address: "0.0.0.0"
consul_rpc_bind_address: "0.0.0.0"
```

Especially as `consul_rpc_bind_address` has already become legacy the more flexible dict object might be the better way to go with this configuration option. Should I open a separate issue to further discuss this topic?

brianshumate:
@oliverprater I think we could go forward with `consul_addresses` if you'd like and feel free to open a new issue for that work.

I'm going to go ahead and merge this now.

Issue #87, opened 2017-06-28
Support Upgrading Consul


oliverprater:
I noticed that when I change the variable `consul_version` and run the script on an existing consul cluster that each nodes still states the old `consul_version` in `consul members`.

When I checked on the machine I noticed that the binaries were **not upgraded** to the specified consul version.

The main issue I found is the **Check for existing Consul binary** task inside the file [task/main.yml](https://github.com/brianshumate/ansible-consul/blob/443d4a5f18c0f00c636d3e5576fd887a4a25e14e/tasks/main.yml#L57).

With the guide from consul at https://www.consul.io/docs/upgrading.html I hacked together something that is not idempotent and uses a variable called `consul_install_upgrade`. This bypasses the stated check and runs additional tasks such as `systemctl daemon-reload` on Ubuntu.

I think further restructuring of this hack is required to ideally make the process idempotent.

So my questions: Is it planned to support Upgrading Consul via the ansible role?

vincent-legoll:
That would be nice to have

brianshumate:
I am definitely not opposed to an upgrade of Consul workflow if we can implement it in a nice way!

oliverprater:
I put in a PR with the described `consul_install_upgrade` workflow, as I think it is not as hacky as I expected it to turn out.

The greatest issue I found, when I started with a solution that calculated the drift between the variable `consul_version` and the output of something like `command: {{ consul_binary }} --version` on the machine was that it might lead to unexpected behavior. While not officially documented, the current behavior is never to upgrade the binaries once installed. A user can, therefore, always pull the newest version of the role without fear that a running consul node will be affected. The `consul_version` variable does not have to be defined in a playbook and as the newer agents are downwards compatible by design the change of the default `consul_version` after a pull might not be noticed. The PR would lead to a breaking change from current behavior!

Worst case scenario is that somebody pulls the newest version of the role with a new version of consul as the default. They don't have a defined version in their playbook. This has worked fine for them as the role never touched installed binaries. They don't notice the breaking change and the running consul cluster goes down because every existing node is subject to an upgrade from which ever version it was running since the first installation.

Finally, it seemed to become a massive mess when trying to handle the described standard rolling upgrade workflow inside the role. The fact that this workflow is not guaranteed to work for each consul upgrade from Hashicrop is another decremental factor.

I don't like variable sprawl myself and am open to discuss any improvements that allow for upgrading consul.

nre-ableton:
[...]

PR #90, opened 2017-07-18
Support Upgrading Consul #87

oliverprater:
This is the described consul_install_upgrade variable, which successfully upgraded consul on systemd.

The guide from hashicorp discusses a rolling upgrade of server and then client nodes, but the atomisation of this inside the role feels wrong.

Note: The Windows code path is not touched by this PR. Somebody with knowledge of that platform should handle it.


brianshumate:
Thank you!

PR #97, opened 2017-08-15
Update config ACLs, Addresses and Ports

groggemans:
Fixes ACL settings and adds a few missing ones.

Adds missing address and port settings and prevents deprecation warnings for rpc settings on nodes using consul > 0.8.0. Not 100% all defaults make sense because the consul docs are a little vague on these and how they compare to each other.

It does have a few non backwards compatible changes to keep the defaults and config template cleaner.
(address and port settings mainly)

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,6 +456,13 @@ ports can be done using the `consul_ports_*` variables.
 - Enable script based checks?
 - Default vaule: *false*
 
+### `consul_raft_protocol`
+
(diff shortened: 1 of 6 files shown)


groggemans:
Made a small mistake in the version compares, fixing this now and also adding the raft_protocol parameter.

brianshumate:
Nice!

PR #99, opened 2017-08-19
dnsmasq updates

groggemans:
Fixes a little bug that prevented dnsmasq from getting the consul DNS data when consul binds to all interfaces (`0.0.0.0`). In this case, dnsmasq will now just query localhost.

Adds the dnsmasq `listen-address` option, only configured if provided.

Updates the dnsmasq service management to make sure it is enabled, and restarted if config is changed.


brianshumate:
👍 Thanks!

PR #110, opened 2017-09-05
Fix list example data type

vincent-legoll:


vincent-legoll:
I've not searched in details, but there may be other instances of this same typo.
And maybe also the datatype is right and the text is wrong, dunno.

PR #113, opened 2017-09-05
Remove debian-only task (uses apt to install packages)

vincent-legoll:


brianshumate:
👍 It does seem that adding python-consul here is pretty much outside the scope of this role.

groggemans:
I do use this in my setup (as the package is needed to use consul from Ansible).

I'm all for keeping it clean and only managing consul itself, but things like this package and the dnsmasque/iptables setup are very handy when building a setup. (consul-template is another nice addition to the list)

No problem if we throw these things out, but it might be nice if we can provide some extra docs, or other roles with examples to replace them?

PR #114, opened 2017-09-05
Fix types that should be lists

vincent-legoll:
Not only in the doc, but also in the defaults.yml file

This is actually untested, and only looked sensible as the text spoke of lists...


vincent-legoll:
I had a quick look at dnsmasq man page and the jinja template, and this ought to be OK, IMHO

groggemans:
Should work just fine. I often use empty dicts where I need empty lists, never had any problems with it. But I'll try to  be a little more consistent!

PR #115, opened 2017-09-05
Fix removed task file include

vincent-legoll:
Fix bug introduced in : fb936053af1663abe4557befb4db36a845043c3a

I tested this is working on a single node install (ubuntu 17.04)


brianshumate:
Our fixes just crossed as I just fixed this as well and made a new release.

PR #125, opened 2017-11-21
Fix idempotency for directories

issmirnov:
The "Create SSL directory" task creates directories as `root:root`, which causes the "Create directories" (from dirs.yml) to always register as changed and thrash the two tasks. Setting ownership to the consul user from the getgo allows us to hit the beautiful `ok=17   changed=0    unreachable=0    failed=0` output on the second run.


brianshumate:
Thanks!

PR #127, opened 2017-11-22
make gossip encryption optional fixes #126

### LINKED ISSUE #126: don't force to encrypt gossip
It would be nice to setup consul client agents to connect to consul servers without encryption

hwmrocker:
I added a new variable `consul_encrypt_enable`.

The default value will not change anything, but if you set it to false it will skip all parts that are handling gossip encryption stuff.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -143,7 +143,7 @@ consul_acl_replication_enable: "{{ lookup('env','CONSUL_ACL_REPLICATION_ENABLE')
 consul_acl_replication_token: "{{ lookup('env','CONSUL_ACL_REPLICATION_TOKEN') | default('', true) }}"
 
 ## gossip encryption
-consul_encrypt_enable: "{{ lookup('env','CONSUL_TLS_ENABLE') | default(true, true) }}"
+consul_encrypt_enable: "{{ lookup('env','CONSUL_ENCRYPT_ENABLE') | default(true, true) }}"
 
 ## TLS
 consul_tls_enable: "{{ lookup('env','CONSUL_TLS_ENABLE') | default(false, true) }}"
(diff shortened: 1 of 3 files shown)


tasks/main.yml
```diff
@@ -79,75 +79,78 @@
...
-      when: consul_raw_key is defined
-
-    # Generate new key if none was found
-    - block:
-        - name: Generate gossip encryption key
-          shell: "PATH={{ consul_bin_path }}:$PATH consul keygen"
-          register: consul_keygen
-
-        - name: Write key locally to share with other nodes
+        - block:
```
hwmrocker, line 83: i just indented the whole part and wrapped a block around it

defaults/main.yml
```diff
@@ -142,6 +142,9 @@ consul_acl_master_token_display: "{{ lookup('env','CONSUL_ACL_MASTER_TOKEN_DISPL
 consul_acl_replication_enable: "{{ lookup('env','CONSUL_ACL_REPLICATION_ENABLE') | default('',true) }}"
 consul_acl_replication_token: "{{ lookup('env','CONSUL_ACL_REPLICATION_TOKEN') | default('', true) }}"
 
+## gossip encryption
+consul_encrypt_enable: "{{ lookup('env','CONSUL_TLS_ENABLE') | default(true, true) }}"
```
brianshumate: Perhaps this Ansible variable can also have its own environment variable, like `CONSUL_ENCRYPT_ENABLE` for example instead of using `CONSUL_TLS_ENABLE`?

brianshumate (changes requested):

hwmrocker:
Oh, 🙈 sure I will fix this.

brianshumate:
Thank you!

PR #131, opened 2017-12-11
extra-args is being ignored because at wrong ident level

chrisparnin:
```
- pip:
    name: bottle
    extra_args: --user
```


groggemans:
thank you!

PR #135, opened 2017-12-29
Feature performance config

t0k4rt:
Hi,
Performance tuning is not available on the current consul conf file template.
I've added this parameter to the template and also reworked the template in order to generate a valid json output.
Could you check this PR and merge it if it's okay for you ?
Thanks !


brianshumate:
👍 Hi, thanks for adding this and for the tests too!

PR #137, opened 2018-01-04
Make gossip key writing idempotent, fixes #136

### LINKED ISSUE #136: Gossip key writing isn't idempotent
My ansible runs are showing the 2 tasks "Write gossip encryption key locally for use with new servers" and "Delete gossip encryption key file" as changing on every Ansible run.  I can't see an obvious conditional to place in the 'when' to prevent them from running.  As it's just a temporary file that's being written to and deleted, is it too much of a lie to put a 'changed_when: false' line in to each task?  I only ask because I get a bit nervous when the summary line shows more things having changed than I initially expected on any given run.

mattburgess:


groggemans:
Thank you!

PR #146, opened 2018-02-13
enhance syslog config, add support for syslog facility and log file

ykhemani:
This PR is related to https://github.com/brianshumate/ansible-consul/issues/144

The CI test fails when the test_results.json is examined by jq, but I believe that that was happening before as well.

Please let me know if you have any questions.

Thanks!
-yash


brianshumate:
Thank you!

Issue #150, opened 2018-02-16
AutoPilot


ChrisMcKee:
Any appetite for adding the config values for autopilot? 
https://www.consul.io/docs/guides/autopilot.html

If there is I can sort it and PR.

brianshumate:
Yeah cool; that'd be pretty nice to have @ChrisMcKee 👍

PR #152, opened 2018-02-21
Add consul autopilot configuration

ChrisMcKee:
Work in progress for #150 autopilot (as I'm yet to run it through test)
I've also added an enterprise attribute as some of the properties are only available in the paid version and I was paranoid about them causing issues in the OSS version (not sure if consulignores or warns)


brianshumate:
Hey @ChrisMcKee,

Thanks for this; I think OSS will error on missing keys for configuration of Enterprise options, so it's good to be cautious about how these are added!

PR #162, opened 2018-04-02
allow modify dns setting

roadrunner:
hello,

with `dns_config`, consul allows [various dns configurations](https://www.consul.io/docs/agent/options.html#dns_config), like `time to live` for node and services.


roadrunner:
just noticed that `consul_config_custom` covers that. closing..

PR #163, opened 2018-04-09
Potential security problem with consul directory permissions

Rtzq0:
Consul directory permissions are world-readable, and may contain ACL tokens.

We could make this configurable, but I thought hard and couldn't come up with a reason to make it readable by other users or the default bin group.


brianshumate:
Thanks!

PR #165, opened 2018-04-30
Clean deprecation warning for version compare

GnomeZworc:
error :
[DEPRECATION WARNING]: Using tests as filters is deprecated.
Instead of using `result|version_compare` instead use `result is version_compare`.
This feature will be removed in version 2.9.


groggemans (approved):
LGTM

Thank you!

PR #168, opened 2018-05-13
Fix consul_configd_path's default value at README

suzuki-shunsuke:
If `consul_config_path` is default value (`/etc/consul`),
`consul_configd_path` is not `/etc/consul.d` but `/etc/consul/consul.d` by default.


brianshumate:
Thanks!

PR #169, opened 2018-05-21
Add Archlinux to supported Linux distributions

huiser:
Add Archlinux to supported Linux distributions


brianshumate:
Thank you!

PR #171, opened 2018-06-08
Add support for environment variables.

coughlanio:
Hashicorp recently added the option to run an updated UI beta in Consul 1.1.0.

https://www.consul.io/intro/getting-started/ui.html#how-to-use-the-new-ui

This needs to be enabled via setting an environment variable, which this playbook had no support for, so I added it anyway. I considered adding a "consul_beta_ui" var, but because this is likely a short lived configuration, I thought it was better to keep it generic.

Thoughts?


brianshumate:
Cool, thanks!

PR #175, opened 2018-07-06
Dependency Management Improvements

logan2211:
- Add a switch to toggle dependency management `consul_install_dependencies`
- Add unzip installation tasks for local/remote installs (toggled by the dependency management switch)
- Fix race condition when installing pip dependency


brianshumate:
Thanks!

PR #176, opened 2018-07-06
Fix the jinja2 retry_join loops

logan2211:
The loops do not aggregate the list of members correctly and the
resulting json only includes the last cluster member in the config.


logan2211:
This may have worked at one point, but on ansible 2.5 and jinja 2.10 I am seeing the broken output. This should probably be tested using a cluster scenario in CI, and I intend to work on testing improvements in a separate PR.

brianshumate:
Looks good, thanks!

PR #182, opened 2018-07-18
Fixed usage of filters as tests (deprecated since 2.5)

viruzzo:
Corrected the usage of filters as tests (i.e. `| changed` ⇒ `is changed`) which generates a warning as of Ansible 2.6 (has been deprecated in **2.5** and is scheduled to be removed in **2.9**).

https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#test-syntax


brianshumate:
Thanks!

PR #186, opened 2018-08-03
Added handler to reload configuration on Linux

viruzzo:
As per https://www.consul.io/docs/agent/options.html, the handler sends the **SIGHUP** signal using `pkill`, which is assumed to be available by default on all Linux distributions supported by the role.

`consul reload` cannot be reliably used because it requires the HTTP API to be bound to a non-loopback interface.

I cannot provide a Windows implementation.

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -7,6 +7,9 @@
 - name: start consul
   import_tasks: start_consul.yml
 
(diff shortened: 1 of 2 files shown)


brianshumate:
Thanks!

PR #187, opened 2018-08-08
fix: generate `consul_acl_master_token` when not provided

abarbare:
Hi!

The `acl.yml` task was never used to generate `consul_acl_master_token` as the variable was present in default variables. 

Removing the variable will generate a new random `uuid` and set it as `consul_acl_master_token`. 

Because of the dependency with the configuration role, it is necessary to launch acl role before the configuration one.


brianshumate:
Thanks!

Issue #189, opened 2018-08-14
cannot install unzip


replicajune:
Hello,

When the role try to install unzip in my setup it fails.

in the install.yml file, at line ~60, the role tries to install unzip, it fails. here is part of the error log :

```
...WARNING: Ignoring APKINDEX.70c88391.tar.gz: No such file or directory\nWARNING: Ignoring APKINDEX.5022a8a2.tar.gz: No such file or directory\nERROR: unsatisfiable constraints:\n", "stderr_lines": ["WARNING: Ignoring APKINDEX.70c88391.tar.gz: No such file or directory", "WARNING: Ignoring APKINDEX.5022a8a2.tar.gz: No such file or directory", "ERROR: unsatisfiable constraints:"], "stdout": "  unzip (missing):\n    required by: world[unzip]\n", "stdout_lines": ["  unzip (missing):", "    required by: world[unzip]"]}
```
The role seems to fail because the package manager was not being asked to be refreshed.

For me, the problem appear because I'm using ansible in a container (based on alpine), and it cannot found unzip because apk is not initialized.

To fix this, some tasks should be set before this one to refresh some packages managers.

I've already identified a couple of module to work with to fix this problem (I've base my research to supported systems) : 
https://docs.ansible.com/ansible/latest/modules/apk_module.html#apk-module
https://docs.ansible.com/ansible/latest/modules/apt_module.html#apt-module
https://docs.ansible.com/ansible/latest/modules/yum_module.html#yum-module
https://docs.ansible.com/ansible/latest/modules/pkgng_module.html#pkgng-module (not sure)

I'm assuming windows wont have this problem because there's not any package manager

if this bug seems to be worth fixing for you I can provide a pr based on tasks that would refresh apk, apt, yum and pkgng.

is it ok ? :smiley:

brianshumate:
Hi there,

Thanks for the report, but is not a bug; this role does not explicitly support Alpine Linux and so it cannot be expected to function properly in that environment.

The distributions which are supported (RHEL/Debian based) and their associated package managers do not exhibit this issue.

If you'd like to do a PR to add the package repository refreshing step and set it to be conditional on the Alpine distribution only, I can merge it in.

replicajune:
rhel and debian can have this issue if references of packages are outdated (should not happen on base repositories though :) )

i'll submit a PR asap with the conditions you asked

replicajune:
Alright,

The first commit didn't work, the second use lookup and fixed the issue.
[...]

PR #191, opened 2018-08-17
Fix apk cache

replicajune:
fixing issue #189 when ansible run on an alpine machine couldn't install unzip because the apk cache wasn't populated nor refreshed


brianshumate:
Thanks!

PR #192, opened 2018-08-18
fix: acl token on existing configuration

abarbare:
This PR allows the user to run the role on an existing configuration without defining `consul_acl_master_token` or `consul_acl_replication_token` by reading configuration file.


brianshumate:
Thanks!

PR #197, opened 2018-08-29
Replace local_action/wait_for with win_wait_for, when install on Windows.

imcitius:
Native `win_wait_for` module works better, and is not relying on dns requests, thus not require hostname to be resolvable.


brianshumate:
Thanks!

PR #198, opened 2018-08-29
Fix windows sha256 read

imcitius:
When current directory is not set to /tmp/consul, findstr cannot open downloaded file with sha256 sums, despite absolute path is mentioned in command.


brianshumate:
Thanks!

PR #199, opened 2018-09-07
Fix undefined is_virtualenv condition

### LINKED ISSUE #161: is_virtualenv treated as defined when defaulted to empty string
Hi All!

Playbook tries to run task "Install netaddr dependency on controlling host (virtualenv)" when varaible VIRTUAL_ENV is not set in VM. Then is_virtualenv is defaulted to '' what is treated as defined value. Then task without user arg is invoked and it results in error caused by lack of permissions on VM.
Connected to [#129](https://github.com/brianshumate/ansible-consul/issues/129)

I'm running role with vagrant 2.0.2 on Win10 against Centos 7 on Virtualbox 5.2.4

jpiron:
As is_virtualenv is part of the defaults variables it is always defined.
Instead it should be checked against empty string or None.

Fixes #161

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -12,7 +12,7 @@
       delegate_to: 127.0.0.1
       become: false
       run_once: true
-      when: is_virtualenv is not defined
+      when: is_virtualenv == '' or is_virtualenv == None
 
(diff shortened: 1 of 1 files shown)


brianshumate:
Thanks!

PR #203, opened 2018-09-13
Ensure idempotent folder permissions

jpiron:
Files and folders created under Consul directories get their permissions
from the umask which is unlikely to be 0700. As a result, each run
changes permissions and report changes. If created files must have specific
permissions, a specific umask should be created for the Consul user.
Moreover, consul_tls_dir is created with 0755 permissions, also leading to
non idempotent runs.
Removing recurse option solve this.


brianshumate:
Thanks!

PR #207, opened 2018-09-30
feat: let user specify systemd restart time

abarbare:
As a user, I want to manage with precision the way consul `systemd` unit works. 
This PR lets the user specify the duration applied to the Consul `systemd` unit in case of failure before trying to restart it. 

A default value is set in order to keep the previous behaviour of the role.


brianshumate:
Thank you!

PR #210, opened 2018-10-30
Fix consul restart handler reference

blaet:
Ansible's handler references (sadly) are case sensitive. This fixes a small bug related to that; trying to restart consul on Linux hosts.


brianshumate:
Thanks!

PR #211, opened 2018-11-01
Write gossip key on all hosts

danielkucera:
if you are not lucky enough the first host won't be the one with consul_raw_key defined

PR #212, opened 2018-11-06
Ensure retrieval of existing consul cluster key

blaet:
Suppose we have a cluster consisting of three consul hosts. The first host has died and has to be newly provisioned.

Setting `run_once: yes` here will run this task only against the facts of that first new host. But that host does not have the encryption key in its config, since it is not configured yet. Therefore, this task will not write anything to the local file system, and the provisioning will silently fail.

Running this task (while delegated to localhost) against all hosts will ensure the `consul_raw.key` will be written to the local file system at least once.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -114,13 +114,13 @@
             - inventory_hostname in consul_servers
 
         # Key provided by extra vars or the above block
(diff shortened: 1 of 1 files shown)


blaet:
Duplicate of #211

PR #215, opened 2018-11-12
Add disable_keyring_file configuration option

vincepii:
Adds the `disable_keyring_file` consul option.

I preferred to omit the line in the configuration file if set to false, rather than including it with a false value, so users will not see this change in their configuration, if they don't need to set the flag.

@brianshumate if you prefer instead to always have the option with the currently configured value in the config file, I can change that.


brianshumate:
Looks good!

PR #217, opened 2018-11-16
Support syslog-ng as well as rsyslog

jeffWelling:
I'd like to add support for syslog-ng as well as rsyslog.

This patch works for me on CentOS7 with syslog-ng 3.5.6.


jeffWelling:
Just a friendly ping here in case this fell through the cracks, I'm very open to making changes if what I've submitted is unsatisfactory - just let me know! :)

brianshumate:
Thanks!

PR #219, opened 2018-11-20
 Do not use grpc configuration before 1.3.0

vincepii:
Grpc configuration on `addresses` and `ports` is not supported on consul versions before 1.3.0.

This PR introduces a check to avoid writing those configuration values when consul version is earlier than 1.3.0.


brianshumate:
Thanks!

PR #223, opened 2018-11-26
Un-gate dnsmasq tasks

### LINKED ISSUE #221: dnsmasq ansible tasks gated by bootstrap status
Hello again, I'm hitting a problem with the dnsmasq portion of the playbook.

**Problem**: The ansible tasks for installing and configuring dnsmasq are gated behind the bootstrap state file existing, meaning that once consul has been deployed for the first time, running the ansible playbook cannot be used to update the dnsmasq config.

**Actual Result**:
Running the ansible playbook to update dnsmasq config after consul has been configured does not result in the new dnsmasq config being deployed.

**Expected Result**:
Running the ansible playbook to update dnsmasq config after consul has been deployed successfully updates the /etc/dnsmasq.d/10-consul file with the new value.

**Proposed solution**:
Unless there is a reason for the dnsmasq config to be gated behind that when condition for bootstrap_state, I propose that the task inclusion be moved below that block.

**References**:
The problem seems to be here: https://github.com/brianshumate/ansible-consul/blob/master/tasks/main.yml#L282

jeffWelling:
Fixes #221


brianshumate:
Thanks!

PR #224, opened 2018-11-26
Un-gate the iptables tasks

### LINKED ISSUE #222: iptables ansible tasks gated by bootstrap status
The iptables ansible tasks are gated by the bootstrap_status variable, similarly to the issue discussed in #221. 

Problem: The ansible tasks for deploying iptables rules are gated behind the bootstrap_status variable, which means ansible rules won't be applied to an already-boostrapped system, even though iptables rules have no relation to bootstrapping.

There's an argument to be made that iptables rules should only be applied to a system while it is being bootstrapped, but I would instead argue that we should trust ansible tasks to be idempotent - an ansible task to create an iptables rule shouldn't execute if that rule already exists. Relying on ansible means we can adjust the iptables rule config tasks and they will still be applied to already-bootstrapped systems.

**Problem:** Running the playbook against an already-bootstrapped host to enable iptables rules doesn't apply any iptables rules.

**Actual Behavior:** Please see problem statement

**Expected Behavior:** Enabling iptables rules via the variable and running the playbook should result in iptables rules being applied

**Proposed Solution:** Move the ansible task which includes the iptables tasks file outside of the block which is gated with the when clause that checks the bootstrap_state. Similar solution to #221

jeffWelling:
Fixes #222


brianshumate:
Thanks!

PR #225, opened 2018-11-27
fix enterprise install consul_pkg name

drewmullen:
consul_pkg should contain '-enterprise' when consul_enterprise = true


brianshumate:
Thanks!

PR #226, opened 2018-11-27
allow for snapshot agent installation and configuration

drewmullen:
to have this playbook enable the consul snapshot agent, set `consul_snapshot` to true and other configurations to your need
such as snapshot dir location
snapshot interval
retention policy
etc

tested and works with TLS enabled and disabled


brianshumate:
Thanks!

PR #229, opened 2018-11-30
[BUG #228] Delegate Consul datacenter DNS domain to Consul.

teralype:
The `dnsmasq` template used to generate `dnsmasq` configuration delegates `{{consul_domain}}` to Consul.

In environments where there are multiple clusters, `dnsmasq` should delegate `{{consul_datacenter}}.{{consul_domain}}`. This way DNS queries to Consul for other datacenters will be routed to upstream DNS servers (which should be configured to support federation).


brianshumate:
Thanks!

danielkucera:
Ping...

brianshumate:
This was made conditional with `consul_delegate_datacenter_dns` with a default of `false` in version 2.3.3 of the role, reverting the behavior to its behavior previous to merging this commit.

Heads up to @teralype and others wishing to delegate the datacenters' domain that you will need `consul_delegate_datacenter_dns: true` with version 2.3.3 and beyond of this role.

PR #230, opened 2018-11-30
Allow dnsmasq to bind to specific interfaces.

teralype:
Enabling this option via `consul_dnsmasq_bind_interfaces` prevents dnsmasq from binding by default 0.0.0.0, but instead instructs it to bind to the specific network interfaces that correspond to the `listen-address` option.


brianshumate:
Thanks!

Issue #236, opened 2019-01-11
Inconsistent permissions on pid file directory


sgrimm-sg:
The role creates (by default) `/var/run/consul` on systemd systems so the PID file can be written there, and sets its permissions to 0700. It also creates a systemd unit file that creates that directory at startup and doesn't set the permissions.

This leads to inconsistent permissions: if a host is rebooted (and thus systemd needs to recreate the run directory) it'll be mode 0755 or whatever the system's default umask dictates. Then when the role is run again, it changes the permissions back to 0700.

brianshumate:
Going to close this since your PR was merged. Thanks!

PR #237, opened 2019-01-11
Set run directory permission in systemd unit file (#236)

sgrimm-sg:
When the run directory is created by systemd at unit start time, set the permissions to match the permissions the role uses when it creates the directory.


brianshumate:
Thanks!

PR #240, opened 2019-01-18
Fix systemd configuration task handler notification

jpiron:
By calling 'restart consul on Linux' instead of 'restart consul' the
Consul server is restarted twice if any other task notifies
'restart consul' as both get executed.


brianshumate:
Looks good, thank you!

PR #242, opened 2019-01-22
Remove token generation/retrieval on clients

jpiron:
ACL master token and replication token are just for Consul servers not
for clients.


brianshumate:
Thanks!

PR #244, opened 2019-01-23
Let retry_join to be set up independently from the hosts servers

ghost:
Let the configuration parameter retry_join to be set at a custom value


brianshumate:
Thanks!

PR #245, opened 2019-01-23
add listen to all the handler tasks

pwae:
In Ansible 2.8, the current way that handlers are done (via import_tasks) will be broken (I found out the hard way!).

See: https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/porting_guides/porting_guide_2.8.rst#imports-as-handlers

This PR makes it so that "listen" commands are added to the tasks themselves, as opposed to the import task defined in handlers/main.yml file.

Tested on:
ansible 2.8.0.dev0 (devel f90ec17465) last updated 2019/01/04 07:54:22 (GMT +800)


brianshumate:
Thanks!

PR #246, opened 2019-01-25
Update default consul version to latest

kkirsche:
https://github.com/hashicorp/consul/blob/v1.4.1/CHANGELOG.md

Available on website now:
https://www.consul.io/downloads.html


brianshumate:
Thanks, but latest is now 1.4.2 and included in the 2.3.2 release of the role!

PR #248, opened 2019-02-06
fix run nssm service on windows

imcitius:
' signs passed directly to consul executable, and prevent consul from opening config files.
Service going to "paused" state, and consul logs file not found error.


brianshumate:
Thanks

PR #249, opened 2019-02-06
fix consul windows handlers

imcitius:
Services in windows cannot be started/restarted this way, due to nssm (not ansible module).
We need to use win_service to do that.


brianshumate:
Okay looks good!

PR #254, opened 2019-02-27
Optional enable old gui

imcitius:
Add ability to enable old GUI style, for those who addicted (like me).


brianshumate:
Thanks!

PR #258, opened 2019-03-20
Fix Windows support for registering services

ghost:
Attempting to register service health checks on Windows would fail due to using ``file`` module without a condition.


ghost:
Actually, the path checking is still off.

brianshumate:
Thank you!

PR #260, opened 2019-03-21
fix consul_bind_address

danielkucera:
when accessing ipv4 in hostvars, dashes in consul_iface need to be converted to underscores
example:
        "ansible_bond_admin": {
            "active": true,
            "device": "bond-admin",
            "features": {
...
            },
            "hw_timestamp_filters": [],
            "ipv4": {
                "address": "10.1.1.1",
                "broadcast": "10.1.1.255",
                "netmask": "255.255.254.0",
                "network": "10.1.1.0"
            },


brianshumate:
Thanks!

PR #271, opened 2019-04-17
Correct snapshot agent config task name

jmariondev:
I imagine this was just a mistake of forgetting to change a copied value, but I noticed this in the file and thought I should PR it.


brianshumate:
Thanks!

PR #277, opened 2019-05-02
Separate OS tasks

### LINKED ISSUE #276: The Individual OS tasks should only be include when needed
The *nix and Windows tasks should be separated out from blocks into individual files that can be included as needed. This will make the code a bit easier to read and can speed up execution of the role in the event you don't have Windows agents, or vice versa.

evilhamsterman:
This simply moves the *nix and Windows blocks into their own files and includes them as needed in the main.yml. This closes #276


brianshumate:
Thanks!

PR #278, opened 2019-05-02
Check if consul_acl_master_token is empty

### LINKED ISSUE #275: ACL tokens will never be automatically generated
The blocks to read and generate the ACL Master and Replication tokens will never execute because the blocks they run in check if `(consul_acl_master_token is not defined)` which is always false because that is defined in defaults/main.yml at an empty string.

evilhamsterman:
Add a check to the ACL tasks to see if the consul_acl_token and the consul_acl_replication_token are empty strings so they can be generated if needed.

This closes #275

### DIFF
--- a/tasks/acl.yml
+++ b/tasks/acl.yml
@@ -16,7 +16,7 @@
 
   when:
     - bootstrap_state.stat.exists | bool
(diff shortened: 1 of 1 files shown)


brianshumate:
Thanks!

PR #282, opened 2019-05-13
Missed an 'or' in replication token check

evilhamsterman:
I missed an or causing the tasks to fail when it tries to run the check


brianshumate:
Thanks!

PR #285, opened 2019-05-16
Fix unarchive failing when "ansible_become:true"

### LINKED ISSUE #284: Consul unarchive fails when "ansible_become:true"
Hello,

I am trying to install Consul using `ansible_become: "true"` and run my playbook using the `ansible-playbook -K` option.

The issue is that when installing, a temporary directory is created so that Consul package can be extracted in it. When running with `ansible_become` this directory is owned by root when created.
The next task, the unarchive does not runs as `root` so the playbook fails because it can't access the folder. 

This issue is concerning this code in `install.yml`

```
- name: Create Temporary Directory for Extraction
  tempfile:
    state: directory
    prefix: ansible-consul.
  become: false
  register: install_temp
  tags: installation
  delegate_to: 127.0.0.1

- name: Unarchive Consul package
  unarchive:
    src: "{{ role_path }}/files/{{ consul_pkg }}"
    dest: "{{ install_temp.path }}/"
    creates: "{{ install_temp.path }}/consul"
  become: false
  vars:
    ansible_become: false
  tags: installation
  delegate_to: 127.0.0.1
```

I have a fix ready for this issue, may I create a PR ?

Thanks and regards

Cyril

Aokrip:
This should fix #284

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -77,6 +77,8 @@
     state: directory
     prefix: ansible-consul.
   become: false
+  vars:
+    ansible_become: false
   register: install_temp
   tags: installation
   delegate_to: 127.0.0.1


brianshumate:
Thank you!

PR #290, opened 2019-06-07
added testing for more distros

gofrolist:
added workaround for ubuntu testing
remove Ubuntu 15.04 (vivid) from meta.yml because it's not LTS release and geerlingguy don't have docker container for it


gofrolist:
I hope you will like these changes.
Now you have testing with molecule against a lot of linux distros:

- centos-6
- centos-7
- debian-8
- debian-9
- ubuntu-16.04
- ubuntu-18.04
- oraclelinux-6
- oraclelinux-7
- fedora-26
- fedora-27
- fedora-28

I removed Ubuntu vivid (15.04) from meta.yml because it's not LTS and there is no available docker image from https://hub.docker.com/u/dokken which I took for testing. But I added Fedora platforms and Debian stretch.

All common parts of molecule configuration I put into molecule/_shared/base.yml and in folders molecule/${SCENARIO}/molecule.yml you have only platform specific configuration.

PR #299, opened 2019-07-17
fix server address for usage with --limit

danielkucera:
this is a workaround for buggy facts gathering when using --limit to deploy only certain hosts
https://github.com/brianshumate/ansible-consul/blob/1606e1ef0fe80307cd5bf37c9c5c243f37086473/tasks/nix.yml#L2


brianshumate:
Thanks!

PR #301, opened 2019-07-25
Updated consul_snapshot.json.j2 to handle TLS correctly and fix line …

jasonneurohr:
…formatting

The current template sets the http_addr to 127.0.0.1:8501 when consul_tls_enable is true. This causes the error below:

> Snapshot agent error: Unexpected response code: 400 (Client sent an HTTP request to an HTTPS server.#012)

To correct this I've added another if check for consul_tls_enable which will append https:// to the front of the consul_client_address value. I also adjusted the whitespace control to format the file.

Before the change the _"ca_file":_ line is appended to the end of the _"http_addr"_ line, as is _"log":_ to the _"key_file":_ line.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -192,6 +192,10 @@ consul_tls_verify_incoming_https: false
 consul_tls_verify_server_hostname: false
 consul_tls_files_remote_src: false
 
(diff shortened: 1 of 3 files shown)


brianshumate:
Thanks!

PR #310, opened 2019-09-05
Fix service restart on upgrades

jpiron:
Systemd module doesn't report changes when only using daemon_reload option.
See: https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/system/systemd.py#L388
result['changed'] is only updated when the unit attribute is provided.
Therefore, the notify statement is never called and Consul is not restarted.


brianshumate:
Thank you!

Issue #311, opened 2019-09-06
Cannot run from Ansible Tower, unzip missing


bbaassssiiee:
In our organization Ansible Tower is managed by another team, and they installed it 'as is', as a generic service. This means that there are no extra's installed for particular teams, like unzip in this case. This breaks the role...

https://github.com/brianshumate/ansible-consul/blob/537108939eda8019c7bbc2975e0b260c4d1ecfd2/tasks/asserts.yml#L98

bbaassssiiee:
Fixed!

PR #313, opened 2019-09-09
🐛 Fix for #311 when unzip is not installed on control host.

bbaassssiiee:
No need to fail when unzip is not on control host. Installing remotely instead.


brianshumate:
Thanks!

PR #314, opened 2019-09-13
Added version check for the log_rotate_max_files key added in consul …

jasonneurohr:
…1.5.3

If the consul version is below 1.5.3 it fails to start because this key is not implemented.

On a host using Consul 1.5.1
[root@consul01 jneurohr]# /usr/local/bin/consul validate /etc/consul/config.json                                                                                             Config validation failed: Error parsing /etc/consul/config.json: 1 error occurred:                                                                                                   * invalid config key log_rotate_max_files

https://github.com/hashicorp/consul/blob/master/CHANGELOG.md#153-july-25-2019
https://www.consul.io/docs/agent/options.html#_log_rotate_max_files


brianshumate:
:doh: good catch, thanks!

PR #317, opened 2019-09-17
Cloud auto discovery

imcitius:
Tag-based support to auto-discovery consul servers in cloud environments, see https://www.consul.io/docs/agent/cloud-auto-join.html

Configuration example:
```
vars:
  consul_cloud_autodiscovery: true
  consul_cloud_autodiscovery_provider: "vsphere"
  consul_cloud_autodiscovery_params: "category_name=consul-role tag_name=consul-server host=vcenter.local user=administrator@vcenter.local password=S3cUR3! insecure_ssl=false"
```


imcitius:
I did not templated `consul_cloud_autodiscovery_params` using simple variables, because parameters differ for different providers.
This approach may need refactoring, but I think it should implement all providers, and result will be very complicated.

brianshumate:
Thanks!

PR #318, opened 2019-09-17
Fix start service on Windows

imcitius:
When on Windows platform, separating create and start service tasks gives better result.
It often fails on first start, if service is not created prematurely.


brianshumate:
Nice, thanks!

PR #320, opened 2019-09-24
Ensure that custom config will not be deleted

slomo:
Specifying `consul_services` will lead to custom config being deleted. In order to prevent this there is a fact that needs to be set


brianshumate:
Thank you!

Issue #315, opened 2019-09-16
Variable acl_master_token never grepped from config.json


bbaassssiiee:
The task **Read ACL master token from previously boostrapped server** from the acl.yml tasks file greps for `acl_master_token` in `config.json`. 

https://github.com/brianshumate/ansible-consul/blob/81d53bdf47215ae9eb97f9d1b504c37cd68f1214/tasks/acl.yml#L8

This string is not present in the config.json of consul 1.6.0 after enabling ACLs, and it seems that **Save acl_master_token from existing configuration** is broken, `ignore_errors: true`

My `config.json` has this structure. The master token is nested in tokens of acl:

```json
{
    "acl": {
        "default_policy": "deny",
        "down_policy": "extend-cache",
        "enabled": true,
        "token_ttl": "30s",
        "tokens": {
            "master": "abcd1234-abcd-ef00-0000-1111222233334444"
        }
    },
```

Parsing json would require an extra library python-jmespath or the `jq` binary. Workaround is to define the acl_master_token with a vaulted variable.

bbaassssiiee:
jmespath is present on Ansible Tower 3.5.0

bbaassssiiee:
This is the correct parser:
 
```yaml
- block:
    - name: Read ACL master token from previously boostrapped server
      shell: "cat {{ consul_config_path }}/config.json"
      register: config_read
      no_log: true
      run_once: true

    - name: Save acl_master_token from existing configuration
      set_fact:
        consul_acl_master_token: "{{ config_read.stdout | from_json | json_query(query) }}"
      vars:
        query: "acl.tokens.master"
      no_log: true
```

PR #321, opened 2019-09-25
🐛 Fix for #315 acl_master_token is not grepped

bbaassssiiee:
acl_master_token is not grepped in Consul 1.6 #315


brianshumate:
Thanks!

Issue #270, opened 2019-04-12
install_remote.yml :: Cleanup broken


vmwiz:
The `with_fileglob` from the Cleanup task of `tasks/install_remote.yml` is incorrect:

```
- name: Cleanup
  file:
    path: "{{ item }}"
    state: absent
  with_fileglob: '{{ /tmp/consul }}'
  tags: installation
```
Or the braces must be removed (like it was before 3cdf0df9b3fdce25a6d0f5d784f49603a864b446)
Or a new variable must be set in the defaults and set accordingly.

PR #324, opened 2019-09-25
🐛 Fix for #270 with_fileglob evaluates a local path.

bbaassssiiee:
image

documentation with_fileglob 
> Matching is against local system files on the Ansible controller. 

https://docs.ansible.com/ansible/latest/plugins/lookup/fileglob.html?highlight=with_fileglob

install_remote.yml can use this simple cleanup.


brianshumate:
Thanks!

Issue #326, opened 2019-09-27
🔐Don't log tokens unless consul_acl_master_token_display is true


bbaassssiiee:
When generating tokens we can set `consul_acl_master_token_display`. When `consul_acl_master_token_display` is false or undefined we should not see `consul_acl_master_token_keygen.stdout` nor `consul_acl_replication_token_keygen.stdout` in the logging.

Please note also: Ideally these tokens differ, but the time-based function allows them to be the same (observed).

bbaassssiiee:
There is a PR to fix this logging exposure.

![image](https://user-images.githubusercontent.com/10671559/65839845-75188080-e311-11e9-846c-965b09c3c03c.png)

PR #327, opened 2019-09-27
🔐 Fix for #326 don't expose the tokens in logging.

bbaassssiiee:
Tokens should not be logged for security reasons. We can accomplish that by use of `no_log: true`. Only because `consul_acl_master_token_display` is set to true for this screen shot we see one token  logged, consul_acl_master_token_display's default value is false.

image


brianshumate:
Thanks!

Issue #325, opened 2019-09-25
Consul 1.4.0+ replication token name changed. replication greps, but this doesn't.


bbaassssiiee:
## replication greps in all versions, but this doesn't.

https://github.com/brianshumate/ansible-consul/blob/81d53bdf47215ae9eb97f9d1b504c37cd68f1214/tasks/acl.yml#L56

https://github.com/brianshumate/ansible-consul/blob/81d53bdf47215ae9eb97f9d1b504c37cd68f1214/templates/config.json.j2#L215

https://github.com/brianshumate/ansible-consul/blob/81d53bdf47215ae9eb97f9d1b504c37cd68f1214/templates/config.json.j2#L240

PR #328, opened 2019-09-27
Fix for #325 for Consul > 1.4.0 Read ACL replication token from previously boostrapped server

bbaassssiiee:
To get the replication token from config.json the grep utility is used with a search string. The problem is that this string changed in Consul 1.4.0 and the search no longer works. This PR has a search string that works in either case.


brianshumate:
Thanks!

Issue #319, opened 2019-09-21
consul_log_path not created for use in log_file or -log-file


bbaassssiiee:
https://github.com/brianshumate/ansible-consul/blob/81d53bdf47215ae9eb97f9d1b504c37cd68f1214/tasks/dirs.yml#L33

bbaassssiiee:
When not logging to syslog the directory for the consul.log is not created

These should be removed for a log_dir to be used:

    - not consul_syslog_enable | bool
    - not consul_configure_syslogd | bool

PR #329, opened 2019-09-27
Fix for #319 Create the log_dir for consul in any case. 

bbaassssiiee:
Log when not setting consul_syslog_enable nor consul_configure_syslogd.


bbaassssiiee:
ignore

Issue #256, opened 2019-03-04
Consul not reloaded when TLS files change


danielkucera:
Consul is not reloaded when TLS files change.

soloradish:
Agents only reload when config file changes. 

https://github.com/brianshumate/ansible-consul/blob/4507528909cfd6a30295f952ce721220514bb605/tasks/config.yml#L26

https://stackoverflow.com/a/35604895/2366058

danielkucera:
yes @soloradish , that is what this issue is about. the same `notify` probably needs to be defined in: https://github.com/brianshumate/ansible-consul/blob/master/tasks/tls.yml

PR #330, opened 2019-09-28
Fix for #256 Consul not reloaded when TLS files change.

bbaassssiiee:
when these files change
```yaml
notify: restart consul
```


brianshumate:
Thanks!

PR #332, opened 2019-10-09
Update windows.yml

FozzY1234-zz:
when using just win_service to create the new service, no chocolatey and nssm is needed


brianshumate:
Thanks!

Issue #253, opened 2019-02-26
flush handlers at the end of the main?


Sispheor:
Hi, 
As handlers are PLAY-scoped and not ROLE-scoped on Ansible. It could be a good practice to add a meta flush handler at the end of the main task. 
So we are sure that the DNS config is up for next roles in the play.

What do you think?

brianshumate:
Hey!

Sounds like a good idea to me!

bbaassssiiee:
@Sispheor can this be closed now?

Sispheor:
Yep.

PR #333, opened 2019-10-10
Flush handlers at the end of the main. Fix for #253.

bbaassssiiee:
As handlers are PLAY-scoped and not ROLE-scoped on Ansible.
It could be a good practice to add a [meta flush handler](https://docs.ansible.com/ansible/latest/modules/meta_module.html)
at the end of the main task. So we are sure that the DNS config is up
for next roles in the play.

https://docs.ansible.com/ansible/latest/modules/meta_module.html

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -52,3 +52,6 @@
   when: consul_services is defined and consul_services|length>0
   tags:
     - consul_services
+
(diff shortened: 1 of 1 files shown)


brianshumate:
Thanks!

PR #339, opened 2019-11-12
sort keys in service configuration

slomo:
This is required to get an idempotent deployment, due to the order of keys in a dict being randomized in python.


brianshumate:
Thanks!

PR #340, opened 2019-11-15
ensure enable_tag_override is json

slomo:
Setting `enable_tag_override` to a boolean in the yaml will result in python printing `True` into the template, which is not valid json.

This change is not backwards compatible for users who were using a service description with before.

```
consul_services:
  - name: name
    ...
    enable_tag_override: "true"
```

What is now supported and in my opinion better:

```
consul_services:
  - name: name
    ...
    enable_tag_override: true
```


brianshumate:
Thanks!

PR #343, opened 2019-11-20
Add 'changed_when: false' to ACL master lookup

jmariondev:
Context: ACLs enabled, using generated tokens

Without this, this resource shows up as changed for every run, even when nothing was changed. This is the same approach used below in this file on the 'Read ACL replication token from...' resource.


brianshumate:
Thank you!

PR #347, opened 2019-11-22
Use generated password as token UUID source

jmariondev:
I'm admittedly not familiar with the internals of "to_uuid", but if it doesn't add any additional randomness source then it might be possible to discover the generated tokens if you knew when the tokens were first generated.

This uses a generated password that [seems to use the OS randomness source](https://github.com/ansible/ansible/blob/4817dcd0fce5cd639e5baf8794bb99e302031739/lib/ansible/utils/encrypt.py#L50) which is more appropriate for a secret token.

### DIFF
--- a/tasks/acl.yml
+++ b/tasks/acl.yml
@@ -23,7 +23,7 @@
 - block:
 
     - name: Generate ACL master token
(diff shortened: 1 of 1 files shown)


brianshumate:
Thanks!

PR #348, opened 2019-11-22
Change capability of consul to bind port < 1024

### LINKED ISSUE #353: Not able to start Consul with port < 1024 when consul_user is not root
Hello,

Currently it's not possible to run consul on port < 1024 due to system limitations.
To avoid this behavior we can use the CapabilityBoundingSet of systemd.

Thanks.
Regards.

smutel:
Hello,

Currently it's not possible to start consul with a non root user.
I added an option in the systemd configuration.

Resolve #353

Thanks.


brianshumate:
Thanks!

Issue #350, opened 2019-11-22
ARM binary change with consul v1.6.2.


KyleOndy:
The release notes [1] for consul 1.6.2 state:

> ARM release binaries: Starting with v1.6.2, Consul will ship three separate versions of ARM builds. The previous ARM binaries of Consul could potentially crash due to the way the Go runtime manages internal pointers to its Go routine management constructs and how it keeps track of them especially during signal handling. From v1.6.2 forward, it is recommended to use:
>
>    consul_{version}_linux_armelv5.zip for all 32-bit armel systems
>    consul_{version}_linux_armhfv6.zip for all armhf systems with v6+ architecture
>    consul_{version}_linux_arm64.zip for all v8 64-bit architectures

This breaks the download on ARM since all the archectures are set to `arm`. [2]

[1] https://github.com/hashicorp/consul/blob/v1.6.2/CHANGELOG.md
[2] https://github.com/brianshumate/ansible-consul/blob/a3dbfc1ac2c60d413bd5b3ad06e5ba48af18a7a1/defaults/main.yml#L9-L21

brianshumate:
Thanks for the heads up and for your PR! I merged your PR and realize it could be a breaking change for previous version users, but would rather the latest version work properly and address when and if it becomes an issue for ARM folks on older versions.

PR #351, opened 2019-11-22
WIP: Support for new ARM builds starting with Consul 1.6.2.

KyleOndy:
Fix for issue #350. Support for new ARM build starting with Consul `1.6.2`.


KyleOndy:
The commit as is works on `armv7l` (RaspberryPi 4).

I am concerned that this change will break for someone using this code base installing a version of Consul prior to `1.6.2`

brianshumate:
Awesome, thanks!

PR #352, opened 2019-11-25
Document consul_manage_user and consul_manage_group in README

karras:
This PR adds the currently undocumented defaults vars `consul_manage_user` and `consul_manage_group` to the README. As it is sometimes handy or desired to change the default group `bin` users might get confused why the group is not automatically created for them. I had to look up the actual task and logic to find out about `consul_manage_group`. I think it makes sense to expose this functionality in the README.


brianshumate:
👍 Cool, thank you!

PR #359, opened 2020-01-07
Encrypt verify connections

arouene:
Add **encrypt_verify_incoming** and **encrypt_verify_outgoing** parameters in the list of customizable parameters.

**encrypt_verify_incoming** - This is an optional parameter that can be used to disable enforcing encryption for incoming gossip in order to upshift from unencrypted to encrypted gossip on a running cluster. See this section for more information. Defaults to true.

**encrypt_verify_outgoing** - This is an optional parameter that can be used to disable enforcing encryption for outgoing gossip in order to upshift from unencrypted to encrypted gossip on a running cluster. See this section for more information. Defaults to true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -565,7 +565,7 @@ Notice that the dict object has to use precisely the names stated in the documen
 ### `consul_encrypt_verify_incoming`
 
 - Verify incoming Gossip connections
-- Default value: false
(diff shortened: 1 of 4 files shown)


brianshumate:
Thank you!

PR #360, opened 2020-01-12
Added GitHub workflows

gofrolist:
[GitHub actions](https://github.com/features/actions) is good CI tool which is released recently. You can have [20 concurrent jobs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#usage-limits) running instead of 5 in Travis.

Migrate from python 2.7 because [Support officially stops January 1 2020](https://www.python.org/dev/peps/pep-0373/#maintenance-releases)
 to python 3.7 with GitHub workflows and python 3.6 with Travis

Added new distros to test against
- CentOS 8
- Debian 10
- Fedora 29
- Fedora 30
- Fedora 31
- Oracle Linux 8 (but it's failed for now)


.github/workflows/galaxy.yml
```diff
@@ -0,0 +1,16 @@
+---
+name: Release to Ansible Galaxy
```
gofrolist, line 2: This file is used to publish Ansible role to Ansible Galaxy.
Same as following code in .travis.yml
```
notifications:
  webhooks: https://galaxy.ansible.com/api/v1/notifications/
```

.github/workflows/galaxy.yml
```diff
@@ -0,0 +1,16 @@
...
+
+on:
+  push:
+    tags:
+      - '*'
```
gofrolist, line 7: It will run only then you push a tag

.github/workflows/galaxy.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: galaxy
+        uses: robertdebock/galaxy-action@1.0.1
+        with:
+          galaxy_api_key: ${{ secrets.galaxy_api_key }}
```
gofrolist, line 16: You will need to create `galaxy_api_key ` in [secrets](https://github.com/brianshumate/ansible-consul/settings/secrets) and put there API Key from Ansible Galaxy https://galaxy.ansible.com/me/preferences

.github/workflows/molecule.yml
```diff
@@ -0,0 +1,44 @@
...
+          - debian-10
+          - fedora-26
+          - fedora-27
+          - fedora-28
+          - fedora-29
+          - fedora-30
+          - fedora-31
+          - oraclelinux-6
+          - oraclelinux-7
+          # - oraclelinux-8
```
gofrolist, line 31: disabled `oraclelinux-8` due to failure with some error regarding systemd
working on fix in docker container

.travis.yml
```diff
@@ -1,27 +1,37 @@
 ---
+dist: bionic
```
gofrolist, line 2: travis is using xenial by default which is quite old and have python 3.5 on system level
here I'm using bionic where we can use python 3.6 with `system_site_packages`
unfortunately python 3.7 is not available with `system_site_packages` option which is required to run `molecule prepare` step

meta/main.yml
```diff
@@ -14,15 +14,23 @@ galaxy_info:
     - name: ArchLinux
       versions:
         - all
+    - name: Debian
```
gofrolist, line 17: Sorted alphabetically by distro name

meta/main.yml
```diff
@@ -14,15 +14,23 @@ galaxy_info:
...
+      versions:
+        - jessie
+        - stretch
     - name: EL
       versions:
         - 6
         - 7
+        - 8
```
gofrolist, line 25: Added version 8 because CentOS 8 passed test successfully (Oracle Linux 8 doesn't)

meta/main.yml
```diff
@@ -14,15 +14,23 @@ galaxy_info:
...
     - name: Fedora
       versions:
         - 26
         - 27
         - 28
+        - 29
+        - 30
+        - 31
```
gofrolist, line 33: Added Fedora 29, 30 and 31 releases.
Maybe we will have only three latest releases and can delete 26, 27, 28?

molecule/_shared/base.yml
```diff
@@ -2,16 +2,12 @@
 scenario:
   test_sequence:
     - lint
-    # - destroy
-    # - dependency
     - syntax
     - create
-    # - prepare
+    - prepare
```
gofrolist, line 7: added prepare step where molecule install required packages (unzip and netaddr)
they should be installed before we run ansible-playbook because it read modules only once while running

molecule/centos-6/molecule.yml
```diff
@@ -6,6 +6,6 @@ platforms:
     groups:
       - consul_instances
     image: dokken/centos-6
-    command: /sbin/init
+    dockerfile: ../_shared/Dockerfile.j2
```
gofrolist, line 9: Dockerfile.j2 moved into _shared folder to make it easier to maintain

molecule/debian-9/molecule.yml
```diff
@@ -7,7 +7,9 @@ platforms:
       - consul_instances
     image: dokken/debian-9
     command: /lib/systemd/systemd
+    dockerfile: ../_shared/Dockerfile.j2
     capabilities:
       - SYS_ADMIN
     volumes:
       - /sys/fs/cgroup:/sys/fs/cgroup:ro
+    privileged: true
```
gofrolist, line 15: some docker images require privilege mode
not sure why

molecule/fedora-26/molecule.yml
```diff
@@ -6,8 +6,9 @@ platforms:
     groups:
       - consul_instances
     image: dokken/fedora-26
-    command: /usr/lib/systemd/systemd
```
gofrolist, line 9: some commands removed if they are the same as in Dockerfile

vars/RedHat.yml
```diff
@@ -2,5 +2,12 @@
 # File: RedHat.yml - Red Hat OS variables for Consul
 
 consul_os_packages:
-  - libselinux-python
+  - "{% if ( ansible_distribution  == 'Fedora' and ansible_distribution_version is version('31', '<') ) or \
```
gofrolist, line 5: This logic added because if OS have python 3 by default it require `python3-libselinux` instead of `libselinux-python`

brianshumate:
Wow this is cool! It's going to take some time for me to try and migrate, but I'll give it a shot.

gofrolist:
> Wow this is cool! It's going to take some time for me to try and migrate, but I'll give it a shot.

You don't need to migrate to GitHub Workflows right now. It can live with .travis.yml simultaneously.
But if you found yourself where you don't look into travis build anymore - you can migrate.

brianshumate:
Thanks!

PR #367, opened 2020-02-04
Enable consul at startup

### LINKED ISSUE #366: Consul is not started automatically at boot
This role does not setup consul to start automatically during the boot process.

smutel:
Fix #366


brianshumate:
Thanks!

PR #369, opened 2020-02-06
Don't create TLS folder if consul_tls_copy_keys is false

smutel:
If consul_tls_copy_keys is false, it means that we already have the TLS certificates installed somewhere. So the creation of the folder in this case is useless.


smutel:
What is missing to merge this PR ?

lanefu (approved):

PR #373, opened 2020-02-13
Use consul_node_name empty by default to use hostname of the current server

paretl:
Related to this issue that I opened yesterday: https://github.com/brianshumate/ansible-consul/issues/371

I think that we can use the default behaviour of Consul to set up the `node_name` but keep the possibility to override it with the Ansible variable `consul_node_name`

The case is that when we build machine images with Consul installed inside, the `node_name` is the hostname of the server used to build the machine image. After that, the node_name will be the same for all VMs created from this machine image and cannot be changed automatically.

What do you think about that?

I'm interested to have feedbacks about it

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,8 +409,10 @@ Notice that the dict object has to use precisely the names stated in the documen
 
 ### `consul_node_name`
 
-- Node name (should not include dots)
-- Default value: `{{ inventory_hostname_short }}`
+- Define a custom node name (should not include dots)
+  See [node_name](https://www.consul.io/docs/agent/options.html#node_name)
+  - The default value on Consul is the hostname of the server.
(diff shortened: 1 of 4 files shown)


lanefu (commented):
This makes sense.

Although it's nice to have consul defaults mirrored in default/main.yml, we shouldn't do it at the expense of hindering functionality.

@paretl have you been able to verify this fixes the problem with your image?

I'll let others weigh in.

paretl:
Hello @lanefu , thanks for your comment. Yes, I'm currently using my fork in production to fix the problem. The hostname is well configured now.

lanefu (approved):

lanefu:
thank you @paretl

PR #374, opened 2020-02-14
Setup the role to be used in check mode

paretl:
As some users are using check mode to be sure that Ansible tasks are correctly done, we can set up the role to work with this mode

For that, we have to ignore alerts on tasks that need a real change on the system (for example move a file that is created in the task before will fail in check mode)

What do you think about that?

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -67,6 +67,7 @@
   when: not consul_package.stat.exists | bool
   run_once: true
   delegate_to: 127.0.0.1
+  ignore_errors: "{{ ansible_check_mode }}"
 
 - name: Update Alpine Package Manager (APK)
   apk:
@@ -97,6 +98,7 @@
     ansible_become: false
   tags: installation
   delegate_to: 127.0.0.1
+  ignore_errors: "{{ ansible_check_mode }}"
 
 - name: Install Consul
   copy:
@@ -108,9 +110,11 @@
   notify:
     - restart consul
   tags: installation
+  ignore_errors: "{{ ansible_check_mode }}"
 
 - name: Daemon reload systemd in case the binaries upgraded
-  systemd: daemon_reload=yes
+  systemd:
+    daemon_reload: yes
   become: true
   when:
     - ansible_service_mgr == "systemd"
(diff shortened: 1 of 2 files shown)


tasks/install.yml
```diff
@@ -108,9 +110,11 @@
   notify:
     - restart consul
   tags: installation
+  ignore_errors: "{{ ansible_check_mode }}"
 
 - name: Daemon reload systemd in case the binaries upgraded
-  systemd: daemon_reload=yes
```
bbaassssiiee, line 113: Thanks for the boy-scouting.

paretl: @bbaassssiiee The value of the PR isn't here. What do you think about other changes? :wink:

bbaassssiiee (changes requested):
I'll test again.

bbaassssiiee (changes requested):
checkmode still fails..
```
TASK [brianshumate.consul : Read ACL master token from previously boostrapped server] ********************************************************************************
Friday 21 February 2020  15:09:49 +0100 (0:00:00.151)       0:00:36.396 *******
skipping: [server01.consul]

TASK [brianshumate.consul : Save acl_master_token from existing configuration] ***************************************************************************************
Friday 21 February 2020  15:09:49 +0100 (0:00:00.391)       0:00:36.787 *******
fatal: [server01.consul]: FAILED! =>
  censored: 'the output has been hidden due to the fact that ''no_log: true'' was specified for this result'
fatal: [server02.consul]: FAILED! =>
  censored: 'the output has been hidden due to the fact that ''no_log: true'' was specified for this result'
fatal: [server03.consul]: FAILED! =>
  censored: 'the output has been hidden due to the fact that ''no_log: true'' was specified for this result'
```

tasks/install.yml
bbaassssiiee: This is a valuable PR, please continue.

paretl:
@bbaassssiiee Thank you for testing and sorry for the long response time
How do you test that the check mode works for every case?
I would like to test it on my machine to add the possibility to launch in check mode for every task

bbaassssiiee:
I use the ansible-consul and ansible-vault roles in https://github.com/dockpack/vault_dojo where I can run these tests in a Vagrant environment on my Mac.

PR #377, opened 2020-02-21
Added support for Alpine

coryjthompson:
Hello, thank you for the great role.

I had to make a few changes to get this role running correctly on Alpine 3.11. 

- Introduced an Alpine variables file 
- Started using OpenRC for the service init scripts. I was initially modifying the sysvinit script however it ended up having so many "if alpine" blocks that it made more sense to introduce an openrc script.

I had to make a call whether consul_run_path was configurable as in the openrc best practices guide it highly recommends against it. I decided to force to /var/run on alpine to keep it standard and added a note to the readme. 

Let me know if I need to make any improvements.

### DIFF
--- a/.github/workflows/molecule.yml
+++ b/.github/workflows/molecule.yml
@@ -14,6 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         scenario:
+          - alpine-3.11
           - centos-6
           - centos-7
           - centos-8
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ virtualenv:
 services: docker
 
 env:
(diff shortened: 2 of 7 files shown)


coryjthompson:
- Github actions is using molecule schema v2 and travis ci is using molecule schema v3.

Details to why Travis build is failing here:
https://molecule.readthedocs.io/en/latest/configuration.html#lint

bbaassssiiee (changes requested):
Please add a `molecule/alpineX` scenario so we can test it. Travis build should pass.

bbaassssiiee:
Travis build should pass.

bbaassssiiee:
Lack of activity.

PR #379, opened 2020-03-01
Molecule 2.22 is used, 3 breaks in CI

bbaassssiiee:
TODO: Upgrade to 3 using the [migration guide ](https://github.com/ansible-community/molecule/issues/2560).


bbaassssiiee:
Having requirements.txt in the root is a good practice.

PR #384, opened 2020-03-30
add leave_on_terminate

ducminhle:
Add `leave_on_terminate` to monitor which consul client or consul server dies if we don't want to leave when machine shutdown. 
https://www.consul.io/docs/agent/options.html#leave_on_terminate


lanefu:
Thank you @ducminhle !

PR #385, opened 2020-03-30
Corrected version Fedora dropped libselinux-python.

jebas:
Fedora switched to python3-libselinux at version 28.


lanefu:
I re-ran the tests and passed on all this time.. . I think the fedora 29 failure was a fluke

lanefu:
thank you @jebas

PR #386, opened 2020-04-01
Allow Role to run when become_method is su

amirhmoradi:
This allows the check for unzip to run when the become_method is set to *su* in a global ansible project and the *su* required a password.


bbaassssiiee:
Do you have a test case or a clear [issue](https://github.com/ansible-community/ansible-consul/issues)? It is unclear what this solves.

In this action we have `become: false`, `ansible_become` is supposed to be equivalent. 

Reference: https://docs.ansible.com/ansible/latest/user_guide/become.html#become-connection-variables

PR #392, opened 2020-05-11
Fix issue 391

### LINKED ISSUE #391: Syslog is used even if log_file is defined
**Step to reproduce:**

- Setup the role with parameters below:
```yml
consul_log_rotate_bytes: 52428800                                                                   
consul_log_rotate_max_files: 31
```

**Description of the issue:**

- The logs files are correctly created
- Syslog event is still setup (logs are sent to /var/log/syslog and /var/log/daemon.log)

smutel:
Fix #391


smutel:
This bug is not related to configuration.
It's related to systemd which grab output of consul to put in syslog.

PR #393, opened 2020-05-11
Stop sending logs to syslog through systemd

### LINKED ISSUE #391: Syslog is used even if log_file is defined

smutel:
Fix #391

### DIFF
--- a/templates/consul_systemd.service.j2
+++ b/templates/consul_systemd.service.j2
@@ -32,6 +32,8 @@ KillMode=process
 KillSignal=SIGTERM
 Restart=on-failure
 RestartSec={{ consul_systemd_restart_sec }}s
+StandardOutput=null
+StandardError=null
 {% for var in consul_env_vars %}
 Environment={{ var }}
 {% endfor %}


lanefu (changes requested):
I feel like these lines should only be added to the unit file when `consul_syslog_enable: true`

could you add jinja conditional ?

smutel:
This is useful in all case:
- When you use consul for log management (consul_syslog_enable: false) => consul will generates logs in his log folder. You don't want to have information twice: in `/var/log/daemon.log` and in `consul_log_path`
- When you use syslog for log management (consul_syslog_enable: false) => consul will use syslog to generate logs. You don't want to have information twice: in `/var/log/daemon.log` and in `consul_log_path`

lanefu (approved):

lanefu:
@smutel Thanks for helping me understand.  merged

PR #398, opened 2020-06-17
Prevent gathering facts for the same servers in loop. Added run_once …

pavel-z1:
If playbook executed for many servers each server interaction will start checking Consul servers facts.

Was added run_once option to prevent not necessary duplicated ssh connections.


bbaassssiiee (approved):
Thanks for spotting this

PR #402, opened 2020-07-23
add auto_encrypt

ducminhle:
Add auto_encrypt to support auto TLS between consul servers and clients.
https://learn.hashicorp.com/consul/security-networking/certificates#client-certificate-distribution
https://www.consul.io/docs/agent/options.html#auto_encrypt


bbaassssiiee (approved):
LFTM, Thanks!

PR #404, opened 2020-08-16
Migrate to molecule V3

gofrolist:
Recently I did change to my molecule-action and rebuild docker image and because of action.yml pointing to latest docker tag it breaks the old version of molecule.

Now workflow uses a versioning for action
`uses: gofrolist/molecule-action@v2.0.0`


bbaassssiiee (approved):
Oh great! Thanks for fixing this.

PR #407, opened 2020-09-03
Update test to generate new encryption key

### LINKED ISSUE #406: Scaling not working with encryption enabled
When we try to scale 3 nodes to 5, ansible read the encryption key on working server and generate a new one for new server.
New server started with a new key and can't join the current cluster.

The issue came from condition of encryption generation's block, which is always true on new node
```yaml
      when:
        - not consul_local_key.changed
        - not bootstrap_state.stat.exists | bool
```

`consul_local_key.changed` is always false because the task has `changed_when` set to false.
The test is also not correct because `consul_local_key` change only on working server not on on new server.

LucasBoisserie:
Current test to determine when generate a new encryption key is always `true` for new server node.

fix #406

### DIFF
--- a/tasks/nix.yml
+++ b/tasks/nix.yml
@@ -89,7 +89,6 @@
       vars:
         ansible_become: false
       no_log: true
-      register: consul_local_key
       delegate_to: localhost
       changed_when: false
       when: consul_raw_key is defined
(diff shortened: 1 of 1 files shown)


bbaassssiiee (approved):
Thanks for the clearly described issue and the elegant fix!

PR #410, opened 2020-09-24
Disable privilege escalation for local zip install check

sgrimm:
If `ansible_become` is set in group or host variables, it takes precedence over task-level `become` settings. Explicitly override it so the local zip install check doesn't require sudo.


bbaassssiiee:
Thanks for the PR with explanation.

PR #411, opened 2020-09-24
Fix service deletion in check mode

sgrimm:
When installing services on a host for the first time, the role fails in check mode because the list of existing service files is not fetched and thus the `list_current_service_config` fact never gets set because it is in a `with_items` loop whose list of items is empty.

Initializing it to an empty list in a separate task means it is set properly even if the list of files is empty.


sgrimm:
Seems like the Molecule failures here are preexisting, nothing to do with this PR, but happy to update my change if needed.

bbaassssiiee:
Failing tests, lack of activity.

PR #417, opened 2020-10-30
systemd: perform the reload before activating the service

marcaurele:
It can fail because the service is not yet seen by systemd. It's safer to perform the reload after adding the service file.


bbaassssiiee:
This looks like a nice fix, but please rebase on master to have the checks working again.

marcaurele:
> This looks like a nice fix, but please rebase on master to have the checks working again.

Done!

PR #418, opened 2020-11-03
fix ansible-lint warnings

gofrolist:
I put `mode `attribute to some files/directories creation where it's not specified


bbaassssiiee (approved):
Thanks for fixing these!

PR #430, opened 2021-01-24
Issue/429 clients wont start on autoencrypt

endyman:
Added If clause to check wether or not we're in auto_encrypt mode and if we render a client config.

### DIFF
--- a/templates/config.json.j2
+++ b/templates/config.json.j2
@@ -91,9 +91,14 @@
     {% endif %}
     {% if consul_tls_enable | bool %}
     "ca_file": "{{ consul_tls_dir }}/{{ consul_tls_ca_crt | basename }}",
+    {% if auto_encrypt is not defined or (auto_encrypt is defined and not auto_encrypt.enabled | bool)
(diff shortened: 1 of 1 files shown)


templates/config.json.j2
```diff
@@ -91,9 +91,14 @@
     {% endif %}
     {% if consul_tls_enable | bool %}
     "ca_file": "{{ consul_tls_dir }}/{{ consul_tls_ca_crt | basename }}",
+    {% if auto_encrypt is not defined or (auto_encrypt is defined and not auto_encrypt.enabled | bool)
+        or (item.config_version != 'client') | bool %}
     "cert_file": "{{ consul_tls_dir }}/{{ consul_tls_server_crt | basename }}",
     "key_file": "{{ consul_tls_dir }}/{{ consul_tls_server_key | basename }}",
     "verify_incoming": {{ consul_tls_verify_incoming | bool | to_json }},
+    {% else %}
+    verify_incoming": false,
```
ahjohannessen, line 100: `"` is missing?

PR #431, opened 2021-01-26
Fix missing comma in config.json.j2

hornwind:
Сorrection of missing comma when specifying consul_limits.


bbaassssiiee:
Thanks!

PR #436, opened 2021-06-01
Consul installation from repository

pavel-z1:
This PR add possibility to install consul from operating system repository, as described here for Linux https://www.consul.io/downloads

To install consul from repository you need to define this variable:
```
consul_install_from_repo: true
```

Changes:
1. consul_configd_path to `/etc/consul.d`, as this is consul documented default directory for services
2. consul_group to `consul`. Reason - should be synced with official configuration in consul repository packages
2. consul_data_path to `/opt/consul`. Reason - should be synced with official configuration in consul repository packages

Supported installation of Consul from repositories for next operating systems:
- CentOS/RedHat/Fedora
- Debian/Ubuntu
- Amazon

Tested on CentOS.

### DIFF
--- a/tasks/install_linux_repo.yml
+++ b/tasks/install_linux_repo.yml
@@ -1,0 +1,105 @@
+---
+# File: install_linux_repo.yml - package installation tasks for Consul
+
+- name: Install OS packages
+  package:
+    name: "{{ item }}"
+    state: present
+  with_items: "{{ consul_os_packages }}"
+  tags: installation
+
+- name: Populate service facts
+  service_facts:
+
+- name: Gather the package facts
+  package_facts:
+    manager: auto
+
+- name: Clean up previous consul data
+  block:
+    - name: Stop service consul, if running
+      systemd:
+        name: consul
+        state: stopped
+      when: ansible_facts.services['consul.service'] is defined
+
+    - name: Remove consul systemd unit file from previous installation
+      file:
+        path: /usr/lib/systemd/system/consul.service
+        state: absent
+      notify: systemctl daemon-reload
+
+    - name: Remove the user 'consul'
+      user:
+        name: consul
+        state: absent
+        remove: yes
+
+  when:
+    - "ansible_distribution|lower == 'redhat' or ansible_distribution|lower == 'centos' or \
+        ansible_distribution|lower == 'fedora' or ansible_distribution|lower == 'amazon' or \
+        ansible_distribution|lower == 'debian' or ansible_distribution|lower == 'ubuntu'"
+    - "'consul' not in ansible_facts.packages"
+
+- name: Install repository
+  block:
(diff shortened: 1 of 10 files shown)


pavel-z1:
@bbaassssiiee can you please review this PR?

tasks/install_linux_repo.yml
```diff
@@ -0,0 +1,103 @@
...
+            ansible_distribution|lower == 'fedora' or ansible_distribution|lower == 'amazon'"
+
+
+    - name: Add an Apt signing key, uses whichever key is at the URL
+      ansible.builtin.apt_key:
+        url: https://apt.releases.hashicorp.com/gpg
+        state: present
+      when: ansible_distribution|lower == 'debian' or ansible_distribution|lower == 'ubuntu'
+
+    - name: Add Redhat/CentOS/Fedora/Amazon Linux repository
```
bbaassssiiee: Please change this to Debian/Ubuntu

bbaassssiiee (changes requested):
Thanks for this!! I like it a lot, you could even consider changing the default to package installation on supported operating systems. Please remove the `ansible.builtin.` prefixes, Ansible 2.9 documents modules without it (2.9 has RedHat support and runs in Tower).

tasks/install_linux_repo.yml
pavel-z1: Updated name to "Add Debian/Ubuntu Linux repository"

pavel-z1:
Hi @bbaassssiiee , I've removed `ansible.builtin.` prefixes.
I didn't change default installation to package because for the same distributives not all versions support installation from repository. If user will need package installation it will enable it.
For now installation by copy binary file will cover more systems by default.

bbaassssiiee (approved):

sandipb:
@pavel-z1 This line seems to have a shell expansion in it: https://github.com/ansible-community/ansible-consul/blob/master/tasks/install_linux_repo.yml#L62

```yaml
repo: "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main"
```

We should probably use `ansible_distribution_release` here instead.

pavel-z1:
Hi @sandipb 
ansible_distribution_release not always return correct data
The problem was observed in previous ansible versions https://stackoverflow.com/questions/61460151/ansible-not-reporting-distribution-info-on-ubuntu-20-04

Therefore used more stable lsb_release at least for now.

sandipb:
@pavel-z1 
I mean shell expansion doesn't work here. It is not interpreted by a shell for it to be expanded.

gardar:
@pavel-z1 
What is the purpose of the `consul_os_prepare_packages`  variables? They don't seem to be used anywhere.

pavel-z1:
Hi @gardar 

Yes, `consul_os_prepare_packages` seems don't used anywhere.
Can be deleted.

PR #437, opened 2021-06-06
distro: add support for Flatcar Linux

ahjohannessen:
- make systemd unit path configurable as
   Flatcar Linux /lib/systemd/system is read-only.

### DIFF
--- a/vars/Flatcar.yml
+++ b/vars/Flatcar.yml
@@ -1,0 +1,6 @@
+---
+# File: Flatcar.yml - Flatcar variables for Consul
+
+consul_os_packages: []
+
+consul_systemd_unit_path: "/etc/systemd/system"
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -52,6 +52,7 @@ consul_manage_group: true
 consul_group: "bin"
 consul_systemd_restart_sec: 42
 consul_systemd_limit_nofile: 65536
+consul_systemd_unit_path: "/lib/systemd/system"
 
 ### Log user, group, facility
 syslog_user: "{{ lookup('env','SYSLOG_USER') | default('root', true) }}"
--- a/tasks/asserts.yml
+++ b/tasks/asserts.yml
@@ -1,11 +1,27 @@
 ---
 # File: asserts.yml - Asserts for this playbook
 
+- name: Define supported *nix distributions
+  set_fact:
+    _consul_nix_distros:
+      - 'RedHat'
+      - 'CentOS'
+      - 'OracleLinux'
+      - 'Fedora'
+      - 'Debian'
+      - 'FreeBSD'
+      - 'SmartOS'
+      - 'Ubuntu'
+      - 'Archlinux'
+      - 'Alpine'
+      - 'Amazon'
+      - 'Flatcar'
+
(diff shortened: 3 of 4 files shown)


ahjohannessen:
@bbaassssiiee WDYT?

bbaassssiiee:
ansible_python_interpreter can best be set in the inventory.
I see too many defines of control_host_ansible_python_interpreter

bbaassssiiee (commented):
Please remove `control_host_ansible_python_interpreter`, you can set it in inventory.

ahjohannessen:
>Please remove control_host_ansible_python_interpreter

@bbaassssiiee  `control_host_ansible_python_interpreter` is set because when host is e.g. MacOS the control host has path `/usr/bin/python` and the target host has, in my case, `/home/core/bin/python` and I have no way of defining a symbolic link to `/usr/bin/python` on Flatcar Linux as `/usr` is read-only.

The problem I am trying to solve with `control_host_ansible_python_interpreter` is that `ansible_python_interpreter` is different on the control host and target host(s) when using Flatcar Linux.

ahjohannessen:
@bbaassssiiee I think I can solve it as you suggest pr. host. I will test and update the PR 👍

ahjohannessen:
@bbaassssiiee I have updated the PR as you suggested! It works fine with overriding `ansible_python_interpreter` in inventory :)

ahjohannessen:
@bbaassssiiee Is there anything else I can do in order to get this merged? :)

bbaassssiiee:
I had to lookup Flatcar Linux, it is supposed to run containers. Why not run consul in a container on it?

ahjohannessen:
> Why not run consul in a container on it?

Mostly because I can piggy-back on the conventions in this role and to avoid running consul itself in a container as it is not recommended by hashicorp.

bbaassssiiee (changes requested):
Please fix `vars/Flatcar.yml` to pass yamllint in the GitHub checks.
Bonus: add a Flatcar scenario to **molecule**

ahjohannessen:
@bbaassssiiee 

>Please fix vars/Flatcar.yml to pass yamllint in the GitHub checks.

Should be fixed now.

>Bonus: add a Flatcar scenario to molecule

I tried to see how the other scenarios are done, but could not find a suitable image for it.

vars/Flatcar.yml
```diff
@@ -0,0 +1,6 @@
+---
+# File: Flatcar.yml - Flatcar variables for Consul
+
+consul_os_packages: []
+
+consul_systemd_unit_path: "/etc/systemd/system"
```
bbaassssiiee: Newline at EOF is missing.

bbaassssiiee (changes requested):

ahjohannessen:
@bbaassssiiee I think it should be ok now.

ahjohannessen:
@bbaassssiiee I have added a list of supported *nix distributions in order to avoid lint error with max characters

bbaassssiiee:
Nice!

bbaassssiiee (approved):

ahjohannessen:
:+1:

PR #439, opened 2021-07-13
install_via_repo: Fix shell expansion using ansible fact and fix versions for apt

sandipb:
The previous apt repo line was using a shell expansion which will not work inside the apt_repository module. Replacing it with a Fact.

### DIFF
--- a/tasks/install_linux_repo.yml
+++ b/tasks/install_linux_repo.yml
@@ -66,7 +66,7 @@
 
 - name: Install consul package
   package:
-    name: "consul-{{ consul_version }}"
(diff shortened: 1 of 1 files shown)


bbaassssiiee:
Thanks for the quick fix!

sandipb:
Another short fix to help consul install on Ubuntu/Debian via repository. When using the `package` module, apt prefers `PKG=VERSION` format instead of `PKG-VERSION`.

asosso:
Thank you @sandipb, it works on Ubuntu 18.04.5 LTS

Issue #429, opened 2021-01-24
When enabling auto_encrypt agent won't start in client mode.


endyman:
When enabling `auto_encrypt` client config contains **cert_file** and **key_file** properties and the agent won't start. The template needs to exclude **cert_file** and **key_file** in clients mode if `auto_encrypt` is enabled.

shinyfoil:
I've also got this problem. I grabbed a copy of your PR, but made a couple changes/fixes to make it more complete.

In the last commit to your fork, you're missing a quote here (https://github.com/endyman/ansible-consul/commit/ad262a02b699966f4e73b5013b990b8cb056aa20#diff-32770839e59ae5da477388c7d674000df855dfcd601cc84471f8362bd24dd8ecR100) that causes the to_nice_json filter to mangle the output.

In the TLS task (https://github.com/ansible-community/ansible-consul/blob/master/tasks/tls.yml#L23-L41), I moved the cert and key tasks into a block that has the same conditional as the config template (except it's using consul_node_role instead of item.config_version) so that it won't break when it can't find the file to copy (which is how I'm using it).

My changes are sitting in a fork right now, but I'm glad to open a PR if someone wants it.

paladin-devops:
@shinyfoil +1 interested in your PR for this fix

PR #441, opened 2021-07-15
Auto Encrypt fix for clients

paladin-devops:
Modified copy TLS files task to break up CA cert and server cert/key, since with auto encrypt the latter two are not required. Also the configuration for those two settings will only be set if auto encrypt is not enabled or if it is enabled and the target is not designated as a client (Consul servers even with auto-encrypt enable do still require those configurations for TLS, if TLS is being used). #429


templates/config.json.j2
```diff
@@ -91,9 +91,14 @@
     {% endif %}
     {% if consul_tls_enable | bool %}
     "ca_file": "{{ consul_tls_dir }}/{{ consul_tls_ca_crt | basename }}",
+    {% if auto_encrypt is not defined or (auto_encrypt is defined and not auto_encrypt.enabled | bool)
+        or (item.config_version != 'client') | bool %}
```
edudobay: Hi @paladin-devops! Nice work, I was looking for a fix for this :)

I have a suggestion to simplify this conditional — the same would apply to the second `when` condition in `tasks/tls.yml`.

```suggestion
    {% if (item.config_version != 'client') or not (auto_encrypt.enabled | default(false) | bool) %}
```

- I put `(item.config_version != 'client')` first to make it more explicit that it doesn't apply to server nodes.
- `(item.config_version != 'client')` is a comparison expression, so casting to `bool` should be unnecessary.
- I rewrote the `auto_encrypt.enabled` part in fewer words, it seems to me easier to understand.

PR #448, opened 2021-12-08
Fix config templating on Ansible 5

rtammekivi:
In Ansible 5, lookup function produces string output (new default? I couldn't find any confirmation about this). Applying to_nice_json provides a double encapsulated JSON output, which is invalid for Consul.

I'd like some feedback as I have only tested this on Linux so far with Ansible 5. Not sure if this is backwards compatible or maybe Ansible introduced a breaking change unindentionally.

Some more info here: https://github.com/ansible/ansible/issues/76443

### DIFF
--- a/tasks/config.yml
+++ b/tasks/config.yml
@@ -6,7 +6,7 @@
     dest: "{{ item.dest }}"
     owner: "{{ consul_user }}"
     group: "{{ consul_group }}"
-    content: "{{ lookup('template', 'templates/config.json.j2') | to_nice_json }}"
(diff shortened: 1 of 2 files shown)


lanefu:
@Blefish   What is Ansible 5?

jlec (approved):
LGTM and works here with ansible-5/ansible-core-2.12

jlec:
@lanefu Ansible-5 is the latest release of Ansible.

lanefu:
hmmm I guess I live too much in the world of Ansible core

jlec:
Thanks for the fix and merging.

PR #460, opened 2022-01-04
Few Improvements to repo install

gardar:
Few improvements to repo install:

- Corrected the repo install prerequisites - the list of packages was the same as used for the archive install and the specific list for the repo install was unused: https://github.com/ansible-community/ansible-consul/issues/436#issuecomment-1000449682
- Made cleanup tasks compatible with other init systems than systemd.
- Simplified the conditionals.
- Added `become` to the tasks that require it
- Specified the apt repo as a variable rather than having it hard-coded (so that it can me overwritten when mirroring the repo internally)

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -28,5 +28,6 @@
   import_tasks: start_snapshot.yml
 
 - name: systemctl daemon-reload
-  ansible.builtin.systemd:
+  systemd:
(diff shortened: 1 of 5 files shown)


gardar:
@lanefu any chance you could merge this? (and perhaps also the other PR's I made :smiley: )

PR #470, opened 2022-04-21
config: fix deprecated usage of ui field.

ahjohannessen:
- use `ui_config` if consul version is 1.9.0 or
   newer.
   See https://www.consul.io/docs/agent/options#ui-1


ahjohannessen:
@bbaassssiiee WDYT?

PR #473, opened 2022-04-22
Fix: Hidden whitespace

ronaldburns:


bbaassssiiee:
Jinja2 filters are separated by pipe characters that are surrounded by spaces.

Issue #469, opened 2022-03-04
dnsmasq regression


zllovesuki:
f5425ddff5853566991e3b72f380384ce021bc4e introduces regression with dnsmasq configuration (target hosts are Debian 11):

```
TASK [consul : Create Dnsmasq configuration] *****************************************************************************************************************************************
changed: [swarm5] => (item={'dest': '/etc/dnsmasq.d/10-consul', 'group': 'root', 'when': 'ansible_os_family|lower != "freebsd"'})
changed: [swarm1] => (item={'dest': '/etc/dnsmasq.d/10-consul', 'group': 'root', 'when': 'ansible_os_family|lower != "freebsd"'})
changed: [swarm0] => (item={'dest': '/etc/dnsmasq.d/10-consul', 'group': 'root', 'when': 'ansible_os_family|lower != "freebsd"'})
changed: [swarm2] => (item={'dest': '/etc/dnsmasq.d/10-consul', 'group': 'root', 'when': 'ansible_os_family|lower != "freebsd"'})
changed: [swarm4] => (item={'dest': '/etc/dnsmasq.d/10-consul', 'group': 'root', 'when': 'ansible_os_family|lower != "freebsd"'})
failed: [swarm5] (item={'dest': '/usr/local/etc/dnsmasq.d/consul.conf', 'group': 'wheel', 'when': 'ansible_os_family|lower == "freebsd"'}) => {"ansible_loop_var": "item", "changed": false, "checksum": "bc7eee12916de28288f07c04d49c6494a6d5e543", "item": {"dest": "/usr/local/etc/dnsmasq.d/consul.conf", "group": "wheel", "when": "ansible_os_family|lower == \"freebsd\""}, "msg": "Destination directory /usr/local/etc/dnsmasq.d does not exist"}
failed: [swarm1] (item={'dest': '/usr/local/etc/dnsmasq.d/consul.conf', 'group': 'wheel', 'when': 'ansible_os_family|lower == "freebsd"'}) => {"ansible_loop_var": "item", "changed": false, "checksum": "bc7eee12916de28288f07c04d49c6494a6d5e543", "item": {"dest": "/usr/local/etc/dnsmasq.d/consul.conf", "group": "wheel", "when": "ansible_os_family|lower == \"freebsd\""}, "msg": "Destination directory /usr/local/etc/dnsmasq.d does not exist"}
[...]

PR #475, opened 2022-05-11
Fix Dnsmasq task

rv32ima:
Fixes the issues that I highlighted in #469. This PR has been tested and is actively deployed.

I'd like a better solution to just flat-out ripping out the FreeBSD check for creating the Dnsmasq configuration though. Would love suggestions!


bbaassssiiee:
@polikarpoff @hatf0 please work out what to do with DNSMasq

PR #477, opened 2022-05-16
Fix directory /usr/local/etc/dnsmasq.d not found

attrib:
When running with ansible 2.15.5 the item.when is not evaluated and is true, therefore it tries to create both configs files. Resulting in a directory not found error.


sandipb:
This is breaking all tests during linting. when statements need to have python conditionals. Jinja expressions doesnt seem to be allowed here.

PR #481, opened 2022-05-30
Improve path handling on Windows

nre-ableton:
This PR fixes a few issues that occur on Windows hosts with regards to path handling.

Primarily, it addresses an issue this role erroneously deleted all the managed service files it creates in the [`Delete not declared services [Windows]`](https://github.com/nre-ableton/ansible-consul/blob/ee6dd334bfeea872271958bfe70c19274c0e2755/tasks/services.yml#L74-L82) task. The source of this problem was mixing POSIX-style and Windows-style paths, so that the file matching [condition](https://github.com/nre-ableton/ansible-consul/blob/ee6dd334bfeea872271958bfe70c19274c0e2755/tasks/services.yml#L78) would always fail and cause the files to be deleted. This PR introduces some special Windows-only tasks that use `win_stat` to convert our internal Unix-style paths to Windows-style paths, which match those discovered by the `win_find` module.

Normally, the Molecule idempotency test would have caught this error, but as Molecule testing on Windows is basically impossible (:crying_cat_face:), it seems to have slipped through the cracks.

This PR also includes similar path conversion tasks when creating the Windows service, and also fixes a bug where a Windows handler was erroneously called from a task with `become: true` (probably [this one](https://github.com/ansible-community/ansible-consul/blob/fa5a9d17066a7bf890ae7623697029c8e7373a59/tasks/install_linux_repo.yml#L87-L101)), which caused an error on Ansible since `become` isn't supported in PowerShell, at least not without extra properties.


nre-ableton:
If https://github.com/ansible-community/ansible-consul/pull/483 is merged, this will likely create a merge conflict with this PR, and it should also be reworked a bit to include `loop_control`/`loop_var` (see 6bd934cdcc135b43ce47d70586e206433918a29a, I'll squash this after review).

nre-ableton:
FYI, I have rebased this branch, squashed, and resolved all merge conflicts. *Edit*: I've also rebased this branch to include the changes in #488 so that it passes CI.

nre-ableton:
Ping @bbaassssiiee

bbaassssiiee:
Please resolve the conflicts (rebase is not possible now)

nre-ableton:
@bbaassssiiee Done!

nre-ableton:
@bbaassssiiee Thanks for the merge!

PR #489, opened 2022-07-12
Improve support for DHCP hosts

nre-ableton:
Using `ansible-consul` with DHCP-configured hosts is problematic, because `0.0.0.0` cannot be used for the `consul_bind_address`, `consul_advertise_address`, or `consul_advertise_address_wan` variables. Since the role determines the IP address when it runs, this will cause problems when a DHCP-based host reboots and has a different IP address than the one written in `config.json`.

To provide a solution for this, this PR introduces new options that allow the user to specify a network interface which will be passed to the service instead.


nre-ableton:
My bad -- this PR isn't necessary. You can also bind to an interface with the role's current configuration by doing:

```yaml
- name: Apply ansible-consul role
  include_role:
    name: "ansible-consul"
  vars:
    consul_advertise_address: >-
      {{ '{{ GetInterfaceIP \"'}}{{ ansible_default_ipv4.interface }}{{'\" }}' }}
    consul_advertise_address_wan: >-
      {{ '{{ GetInterfaceIP \"'}}{{ ansible_default_ipv4.interface }}{{'\" }}' }}
    consul_bind_address: >-
      {{ '{{ GetInterfaceIP \"'}}{{ ansible_default_ipv4.interface }}{{'\" }}' }}
```

Sorry for the noise!

PR #488, opened 2022-07-12
Fix CI build errors

nre-ableton:
This PR contains a few fixes for CI errors, which mostly resulted from PRs being merged without first being rebased and re-tested.

PR #490, opened 2022-07-12
Add Mac OS X support

### LINKED ISSUE #486: Support for macOS
It would be great if this role could support macOS hosts. consul itself supports macOS (also on Apple Silicon), so this shouldn’t be too difficult, considering most of the tasks are already in place.

As far as I can tell, the only thing that is necessary is to define Mac-specific vars for the role and provide a launchctl plist file. I actually have a [branch](https://github.com/ansible-community/ansible-consul/compare/master...nre-ableton:ansible-consul:nre/master/mac-support?expand=1) which implements this that we have been testing on Mac hosts in our cluster.

If there are no strong objections to adding support for this platform, I’ll go ahead and submit a PR. I’ll wait one week for responses before doing so (so on 14/07/22).

nre-ableton:
Fixes #486. Note that this branch also contains the changes in #488 so that it passes CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ The role might work with other OS distributions and versions, but is known to fu
 * CentOS: 7, 8
 * Debian: 9
 * FreeBSD: 11
+* Mac OS X: 10.15 (Catalina)
 * RHEL: 7, 8
 * Rocky Linux: 8
 * OracleLinux: 7, 8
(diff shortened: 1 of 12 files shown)


bbaassssiiee (approved):
Nice work!

nre-ableton:
@bbaassssiiee Thanks for the merge!

Issue #466, opened 2022-02-14
33b0f4d44e39717bf8cf65d83fb0411e0f7f0892 introduces undefined variable consul_os_repo_prerequisites


github-abcde:
This commit introduces a variable `consul_os_repo_prerequisites` in `tasks/install_linux_repo.yml`, however the variable name in `vars/(Amazon|Debian|RedHat).yml` is named `consul_repo_prerequisites` (if that is the correct variable, otherwise there is none), causing ansible to abort in the `Install OS packages`-task with a `consul_os_repo_prerequisites is undefined`-error.

PR #491, opened 2022-07-15
Fix undefined variable for repo installs

sandipb:
https://github.com/ansible-community/ansible-consul/pull/460 introduced an typo causing repo installations to fail

```
TASK [ansible-consul : Install OS packages] ***********************************************************************************************************************
fatal: [neptune]: FAILED! => {"msg": "The conditional check '(consul_os_repo_prerequisites)' failed. The error was: error while evaluating conditional ((consul_os_repo_prerequisites)): 'consul_os_repo_prerequisites' is undefined\n\nThe error appears to be in '***/.cache/roles/ansible-consul/tasks/install_linux_repo.yml': line 4, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Install OS packages\n  ^ here\n"}
```

Also reported here: https://github.com/ansible-community/ansible-consul/issues/466

### DIFF
--- a/tasks/dnsmasq.yml
+++ b/tasks/dnsmasq.yml
@@ -37,7 +37,7 @@
     mode: 0644
   become: true
   notify: restart dnsmasq
-  when: "{{ item.when }}"
+  when: item.when
   tags: dnsmasq
   loop:
     - { dest: '/etc/dnsmasq.d/10-consul', group: 'root', when: ansible_os_family|lower != "freebsd" }
--- a/tasks/install_linux_repo.yml
+++ b/tasks/install_linux_repo.yml
@@ -6,7 +6,7 @@
     name: "{{ consul_repo_prerequisites }}"
     state: present
   become: true
-  when: (consul_os_repo_prerequisites)
+  when: (consul_repo_prerequisites)
   tags: installation
 
 - name: Populate service facts


tasks/dnsmasq.yml
```diff
@@ -37,7 +37,7 @@
     mode: 0644
   become: true
   notify: restart dnsmasq
-  when: "{{ item.when }}"
+  when: item.when
```
nre-ableton, line 40:
```suggestion
  when: "{{ dnsmasq_item.when }}"
```

See #488.

sandipb: @nre-ableton 
The lint error was about jinja expressions being used in `when`. It only expected pure python expressions. Have you tried this out by lint?

nre-ableton: @sandipb True, but there are several other places in this role where such expressions are used, which is why my PR disabled the expression for the time being. I think that changing the other usages and re-enabling the warning deserves a separate PR.

However, my suggestion is not so much about the Jinja2 expressions but rather the use of `item`, which will be undefined and fail. That was named to `dnsmasq_item` [here](https://github.com/ansible-community/ansible-consul/commit/3cb15f3de7a6b29e37d55f90f52e446141af19d8#diff-6c7ae99af38f268480961efb10a100f3c9a521c6cbeb0f6cab760ad42d1f625aR45-R46).

If you prefer, you could also do:

```suggestion
  when: dnsmasq_item.when
```

bbaassssiiee: Duplicate

PR #496, opened 2022-08-08
Allow to add telemetry

AlbanAndrieu:
Allow to add telemetry options and override them for prometheus

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -256,6 +256,11 @@ consul_services: []
 # enable Consul Connect
 consul_connect_enabled: false
 
+# enable Consul Telemetry
+consul_telemetry_enable: false
+consul_telemetry_prometheus_retention_time: "48h"
+consul_telemetry_disable_hostname: true
+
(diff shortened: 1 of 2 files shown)


abarbare:
Hello, 
You can use the `consul_config_custom` variable which will create a `consul.d/50custom.json` that will be merged with the config.json file and let you add any option you want even if they are not defined in the `config.json.j2` 

i.e
```
consul_config_custom:
    telemetry:
        prometheus_retention_time: "48h"
        disable_hostname: true
```

AlbanAndrieu:
> Hello, You can use the `consul_config_custom` variable which will create a `consul.d/50custom.json` that will be merged with the config.json file and let you add any option you want even if they are not defined in the `config.json.j2`
> 
> i.e
> 
> ```
> consul_config_custom:
>     telemetry:
>         prometheus_retention_time: "48h"
>         disable_hostname: true
> ```

Thanks @abarbare for the advice I will close the pull request then

PR #499, opened 2022-08-18
config: account for deprecated settings in 1.11.0

ahjohannessen:
@bbaassssiiee WDYT?

See https://www.consul.io/docs/upgrading/upgrade-specific#consul-1-11-0

I have fixed dnsmasq.yml with respect to an ansible-lint error.


bbaassssiiee:
LFTM

PR #507, opened 2022-10-04
Support upgrading consul when version changes

### LINKED ISSUE #87: Support Upgrading Consul

nre-ableton:
This PR introduces a new variable, `consul_force_install`, which when true, will forcibly install Consul on the host. Otherwise, when this role runs, the Consul version is inspected with `consul --version` and compared to `consul_version` to determine if an upgrade is necessary. Thus, the `consul_install_upgrade` variable is no longer necessary and has been removed by this PR.

This PR also contains a few small cleanups and fixes, primarily for Windows (which I suspect is somewhat less rigorously tested than other platforms).

I have tested these changes locally to make sure that initial installation, idempotent installation, and an upgrade all work correctly. I have tested on the following platforms:

- Linux (Ubuntu 20.04)
- Mac 10.15
- Windows 10

Fixes https://github.com/ansible-community/ansible-consul/issues/87

### DIFF
--- a/README.md
+++ b/README.md
@@ -666,6 +666,13 @@ auto_encrypt:
   ip_san: ["127.0.0.1"]
 ```
 
+### `consul_force_install`
+
+- If true, then always install consul. Otherwise, consul will only be installed either if
+  not present on the host, or if the installed version differs from `consul_version`.
+- The role does not handle the orchestration of a rolling update of servers followed by client nodes
+- Default value: false
+
 ### `consul_install_remotely`
 
 - Whether to download the files for installation directly on the remote hosts
--- a/README.md
+++ b/README.md
@@ -679,13 +679,6 @@ auto_encrypt:
 - This is the only option on Windows as WinRM is somewhat limited in this scope
 - Default value: false
 
-### `consul_install_upgrade`
-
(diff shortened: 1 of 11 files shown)


bbaassssiiee:
Thanks!

nre-ableton:
@bbaassssiiee Thank _you_ for the reviews and merges!

PR #514, opened 2022-11-15
Improve linting

ppuschmann:
@bbaassssiiee Another try...  ;-)


tasks/asserts.yml
```diff
@@ -112,7 +112,7 @@
 # Check for unzip binary
 
 - name: Check if unzip is installed on control host
-  shell: "command -v unzip -h >/dev/null 2>&1"
+  ansible.builtin.shell: "command -v unzip -h >/dev/null 2>&1"
```
bbaassssiiee, line 115: CRITICAL Error see build log

bbaassssiiee (changes requested):
tasks/asserts.yml action starting on line 114 is unparsable.

ppuschmann:
Let me have look at this tomorrow.

tasks/asserts.yml
vitabaks: `ansible.builtin.` (collection) available starting from ansible 3 (2.10).
And this means that there will be no compatibility with ansible 2.9 which is still used by some distributions (e.g. debian 10, ubuntu 20.04)

@ppuschmann Consider ignoring this warning and leaving compatibility with older versions of Ansible.

ppuschmann: @vitabaks Thank you for reminding me on this PR.

So I should rather drop the commits d011257c4f264a9a2b4b8d9863177fb0da258705 and 5b6645b4ebadc5c285354739020660e83f943827 and maybe add some ignore-pattern to `.ansible-lint.yml`?

bbaassssiiee: Rather re-work the PR to do a remote install as the unzip won't work in an execution environment (AWX/AAP)

ppuschmann: I don't quite get the issue.

Is there a general issue with with this unzip? 8and maybe limited to AWX/AAP?
-> I don't really want to touch this right now.

ppuschmann: I think I'll just close this PR for now and think about re-opening another one or multiple smaller ones.

Especially after also stumbling over specialties of `yamllint` requiring "two" spaces before a comment. 
https://github.com/redhat-developer/vscode-yaml/issues/433

Sorry.

PR #528, opened 2023-02-24
Update molecule-action for GitHub workflows

nre-ableton:
This PR updates the molecule-action GitHub action from 2.0.1 -> 2.3.13. This is a pretty big change, since it also implies that we are updating the following:

- Molecule: 3.0.8 ➡ 4.0.4
- Ansible-lint: 4.3.3 ➡ 6.13.0

Because ansible-lint in particular has introduced a _lot_ of new rules since 4.3.x, I've suppressed any new violations for the time being but with the intention that they'll be cleaned up in the near-ish future.


nre-ableton:
I've added @bbaassssiiee as a reviewer, but I also welcome any reviews or input from the general community. I understand that @bbaassssiiee is quite busy and doesn't have a lot of time for this repository, so if there are no opposing comments, I'll merge this PR in a week (on March 3rd).

bbaassssiiee:
Ensure to 'Rebase and merge' pull-requests...

molecule/_shared/base.yml
```diff
@@ -21,6 +21,7 @@ lint: |
   yamllint .
   ansible-lint
   flake8
+role_name_check: 1
```
bbaassssiiee, line 24: Will that work?

tests/test.yml
```diff
@@ -1,5 +1,6 @@
 ---
-- hosts: localhost
+- name: Test play
```
bbaassssiiee: I don't think this tests directory is needed

molecule/_shared/base.yml
nre-ableton: Yes, see:

https://molecule.readthedocs.io/en/latest/configuration/#role-name-check

Unfortunately it is necessary because this role's name doesn't comply with the new Ansible naming standards. At some point in time, it would be nice if we could rename the role, but that's a much larger discussion and probably also one that should be taken up with other elements of the Ansible community.

tests/test.yml
nre-ableton: Great! I wasn't 100% sure about that. Anyways, removed in 541fd5e.

molecule/_shared/base.yml
bbaassssiiee: Ever since moving this into this organization it couldn't be published to Galaxy. Hashicorp doesn't want to spend resources on this. Maybe the namespace in meta/main.yml can be changed to comply. You could maybe setup another organization?

nre-ableton: Right, yes, I recall that now. I am actually happy with the role being under the ansible-community org, and moving it might be disruptive for users. I'll have a look at what could be done for the namespace. Again, this is part of a bigger discussion, and if I find a solution it'll be fixed in a separate PR.

bbaassssiiee (approved):
Lot's of changes in one PR, but the tests pass. WCGW?

nre-ableton:
@bbaassssiiee Thanks for the reviews. :+1:

PR #542, opened 2023-03-19
Add ansible-lint ingnore rule "fqcn-builtins"

ppuschmann:
Ignore the hint to use fqcn-names even for builtin modules.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -4,6 +4,7 @@ skip_list:
   - "key-order[task]"
   - "fqcn[action-core]"
   - "fqcn[action]"
+  - "fqcn-builtins"
   - "ignore-errors"
   - "jinja[spacing]"
   - "no-changed-when"


nre-ableton (changes requested):
Why was this suppression added? Are you using a different version of ansible-lint than the project? It should be at 6.13.1, which is the same version that CI is using. I'm aware that the `requirements.txt` file lists something different, I'll push a PR to fix that.

ppuschmann:
The supression was added because someone wanted to keep supporting older Ansible-versions, IIRC.

nre-ableton:
FQCN support was [introduced in Ansible 2.5](https://www.ansible.com/blog/ansible-2.5-traveling-space-and-time), and this role requires that same version as the [minimum Ansible version](https://github.com/ansible-community/ansible-consul/blob/da647f90aeda735ca25b76fe59d8f453455afda4/meta/main.yml#L8). So we should be able to assume that the Ansible version supports FQCNs and (IMHO) there is no reason to add this suppression.

PR #547, opened 2023-04-07
Fix gossip key generation failures, CI

### LINKED ISSUE #506: Task "Write gossip encryption key locally for use with new servers" is flakey
We often see this task fail in our CI runs, but since it has the `no_log` property, we don't have logging output that might be useful in helping to figure out why.

Is anyone else experiencing this? Also, it might be nice to have a var to control logging output for such tasks rather than assuming `no_log: true`.

nre-ableton:
The primary purpose of this PR is to fix the gossip [key generation issues](https://github.com/ansible-community/ansible-consul/issues/506), which this PR does by avoiding the race condition in Ansible when using `copy` with the `content` attribute when delegated to `localhost`.

Additionally, I "fixed" CI by suppressing `risky-shell-pipe` in ansible-lint. Unfortunately, adding `set -o pipefail` ahead of the `shell` commands isn't a great solution because the Molecule tests execute with `/bin/sh`, which unlike `/bin/bash` lacks support for `set -o pipefail`. The version of the shell that ships with the Docker container used by the Molecule tests also doesn't support `set -o pipefail`, so setting Molecule's executable won't work either.

FWIW, the reason that CI is currently broken is because even though _we_ pinned the `molecule-action` GitHub action, _it_ doesn't pin the Docker image correctly (see https://github.com/gofrolist/molecule-action/issues/140). An update to this image now forces us to fix new ansible-lint violations just to get CI back up and running.

Fixes https://github.com/ansible-community/ansible-consul/issues/506

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -13,6 +13,12 @@ skip_list:
   - "schema[meta]"
   - "var-naming"
 
+  # Pipefail is a bit more difficult, because the shell shipped in the Docker image used
+  # by the Molecule tests doesn't recognize the pipefail option. Likewise, we also need to
+  # make sure that the playbook works for hosts that don't have a modern bash or a default
+  # shell such as /bin/sh.
+  - "risky-shell-pipe"
+
   # These errors should continue to be ignored
   - "no-jinja-when"
   - "role-name"
--- a/tasks/asserts.yml
+++ b/tasks/asserts.yml
@@ -100,8 +100,7 @@
 - name: Fail if more than one bootstrap server is defined
   fail:
     msg: "You can not define more than one bootstrap server."
-  when:
-    - _consul_bootstrap_servers | length > 1
+  when: _consul_bootstrap_servers | length > 1
 
 - name: Fail if a bootstrap server is defined and bootstrap_expect is true
   fail:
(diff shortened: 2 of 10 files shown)


bbaassssiiee:
`molecule.yml` can embed ansible configuration to allow for a shell supporting pipefail (provided it's installed):
```yaml
provisioner:
  name: ansible
  config_options:
    defaults:
      executable: /bin/bash

```

nre-ableton:
> `molecule.yml` can embed ansible configuration to allow for a shell supporting pipefail (provided it's installed):
> 
> ```yaml
> provisioner:
>   name: ansible
>   config_options:
>     defaults:
>       executable: /bin/bash
> ```

Thanks for the advice, I didn't know about this configuration trick. I did some experimentation, but unfortunately I couldn't get it to work because the `dokken` Docker images ship a bash version that doesn't support `set -o pipefile`.

As such, I've updated the commit messages, comment in `.ansible-lint`, and PR description. Please have another look when you have a moment @bbaassssiiee, thanks!

bbaassssiiee (approved):
Thanks!

nre-ableton:
@bbaassssiiee Thanks for the review!

PR #549, opened 2023-04-21
Improve Molecule testing scenarios

nre-ableton:
This PR removes scenarios for obsolete OS versions, and adds support for the currently-supported versions of other OS's that we were already testing.

Even though Ubuntu 18.04 is still _technically_ supported, that will end in about a month or so, so I think it makes sense to end it now rather than creating another PRs.


bbaassssiiee (approved):
Nice one

PR #541, opened 2023-03-19
Improvements for `var-spacing` in ansible-lint

ppuschmann:
The rule was formerly known as `jinja[spacing]`.

PR #550, opened 2023-04-21
Run ansible-lint in its own workflow

nre-ableton:
This is in preparation for Molecule 5.x, which removes the `lint` command. For more information, see:

https://github.com/ansible-community/molecule/pull/3802

Aside from that, this approach is also a bit nicer since linter errors can be identified by their own GitHub checks run, which should be easier for developers compared to the previous approach of having to hunt through Molecule's error logs for linter errors. Also, it's a bit more efficient since we only run the linters once per CI run instead of in every Molecule scenario.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,6 +1,7 @@
 ---
 skip_list:
   # TODO: Fix ansible-lint violations for these errors and remove from skip_list
+  - "fqcn-builtins"
   - "fqcn[action-core]"
   - "fqcn[action]"
   - "ignore-errors"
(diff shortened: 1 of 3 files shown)


nre-ableton:
Note: After https://github.com/ansible-community/ansible-consul/pull/541 is merged, we should be able to remove `var-spacing` that was added in 8aa884623366c5a097b6f9b81f4bef837e1500b4 for the purposes of making CI pass.

bbaassssiiee:
We need another maintainer to review?

nre-ableton:
> We need another maintainer to review?

No, you just need to approve and I'll merge

bbaassssiiee (approved):
Looks good to me

nre-ableton:
> Looks good to me

Thanks!

PR #551, opened 2023-04-25
Ensure tempfiles are defined before deleting them

nre-ableton:
In some cases, if the block exits early then this variable will be undefined, which would cause a failure.


bbaassssiiee (approved):

nre-ableton:
@bbaassssiiee thanks for the merge!

PR #552, opened 2023-04-28
More improvements to gossip key handling

nre-ableton:
Sorry, I know there have been a few PRs already on this topic, but I hope that this is the last one for awhile. 😊 This PR contains two major changes:

1. Removal of the long-obsolete `encrypt-gossip.yml` tasks file. This file is only referenced in a comment, and hasn't actually been used for some years. It's presence is confusing and it has unwittingly received regular updates, particularly because of ansible-lint forcing us to maintain it. I suppose there is a slim chance that people are directly invoking this particular tasks file from this role, but not only is that unlikely, it should be discouraged since this code is effectively untested.
2. Use of `tempfile` rather than the hardcoded `/tmp/consul_raw.key` file. As mentioned in the commit message of ecda07f0f8fe2dfbd8d00cbd2edb7ae19047b37f, this helps to avoid a potential security issue as well as a race condition that could be provoked when running multiple playbooks with this role. On the subject of parallelism, there's also a commit to add a `run_once` attribute to a task to avoid potential race conditions within a single playbook instance.

This PR also contains a few smaller fixes that shouldn't affect the overall behavior.


nre-ableton:
@bbaassssiiee Thanks! :+1:

PR #553, opened 2023-05-15
Update README.md

avalor1:
Default is set to TLSv1_2 but README said tls12.
Also the documentation and changelogs are not quite clear about when this values are used so i added another point for this after comparing different tags of consul. Seems to have change with 1.11.11 to 1.12.0


nre-ableton (approved):
Thanks!

PR #560, opened 2023-07-17
Update molecule-action to 2.6.10

nre-ableton:
2.3.13 was released on Feb 18, 2023, so we're pretty out-of-date. Fortunately everything still seems to work for CI.


bbaassssiiee (approved):
LFTM

PR #564, opened 2023-08-04
Specify virtual env path when using a virtual env.

bastienwirtz:
In order to properly load the environment when using a virtual env, the `virtualenv` [parameter](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pip_module.html#parameter-virtualenv) must be specified. 

This patch fix the the `externally-managed-environment` error ([since PEP 668 has been adopted recently](https://pythonspeed.com/articles/externally-managed-environment-pep-668/)): 
```ansible 
TASK [ansible-consul : Install netaddr dependency on controlling host (virtualenv)] *****************************************************************************************************************************************************************************************************
fatal: [consul-lab01 -> 127.0.0.1]: FAILED! => {"changed": false, "cmd": ["/usr/bin/python", "-m", "pip.__main__", "install", "netaddr"], "msg": "\n:stderr: error: externally-managed-environment\n\n× This environment is externally managed\n╰─> To install Python packages system-wide, try 'pacman -S\n    python-xyz', where xyz is the package you are trying to\n    install.\n    \n    If you wish to install a non-Arch-packaged Python package,\n    create a virtual environment using 'python -m venv path/to/venv'.\n    Then use path/to/venv/bin/python and path/to/venv/bin/pip.\n    \n    If you wish to install a non-Arch packaged Python application,\n    it may be easiest to use 'pipx install xyz', which will manage a\n    virtual environment for you. Make sure you have python-pipx\n    installed via pacman.\n\nnote: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.\nhint: See PEP 668 for the detailed specification.\n"}
```

Also, `is_virutalenv` was renamed to `virtualenv`, as it's not a boolean (could also be `virtualenv_path` and moved to the path section of the variable defaults?)


defaults/main.yml
```diff
@@ -3,7 +3,7 @@
 
 ## Core
 consul_debug: false
-is_virtualenv: "{{ lookup('env','VIRTUAL_ENV') | default('', true) }}"
+virtualenv: "{{ lookup('env','VIRTUAL_ENV') | default('', true) }}"
```
nre-ableton:
```suggestion
virtualenv: "{{ lookup('env','VIRTUAL_ENV') | default('') }}"
```

tasks/main.yml
```diff
@@ -18,17 +18,18 @@
       vars:
         ansible_become: false
       run_once: true
-      when: not is_virtualenv or is_virtualenv == None
+      when: not virtualenv or virtualenv == None
```
nre-ableton:
```suggestion
      when: virtualenv == ""
```
...since it will always be defined in the role defaults.

tasks/main.yml
```diff
@@ -18,17 +18,18 @@
...
       pip:
         name: netaddr
+        virtualenv: "{{ virtualenv }}"
       delegate_to: 127.0.0.1
       become: false
       vars:
         ansible_become: false
       run_once: true
-      when: is_virtualenv is defined
+      when: virtualenv is defined
```
nre-ableton:
```suggestion
      when: virtualenv != ""
```

Again, it will always be defined in the role defaults.

nre-ableton (commented):
Thanks for the PR. :+1: Sorry I missed this one while reviewing the others.

bastienwirtz:
No worries @nre-ableton, thanks for the review !

nre-ableton (commented):
LGTM, but can you please rebase and fix conflicts? You might need to fix some linting issues as well since we've disabled a bunch of ansible-lint suppressions.

bastienwirtz:
Updated @nre-ableton

nre-ableton (approved):
LGTM, I'll just wait for CI to finish before merging.

Issue #518, opened 2022-12-22
Support for consul validate


Lusitaniae:
Ideally we'd have some validation task when updating template files

I'm trying something like this
`
    # validate: "consul validate %s {{ consul_configd_path }}"
`

Which fails with:
```
Config validation failed: Multiple private IPv4 addresses found. Please configure one with 'bind' and/or 'advertise'.
```

Related to  https://github.com/hashicorp/consul/issues/3897?

bastienwirtz:
Hello @Lusitaniae, 

Have you tried adding the main configuration file in the same command ? 
```sh
consul validate {{ consul_config_path }}/config.json {{ consul_configd_path }}
```
I just opened a PR to add the configuration validation (cf https://github.com/ansible-community/ansible-consul/pull/565).

PR #565, opened 2023-08-04
Add support for configuration validation

bastienwirtz:
This patch add an additional step to  ensure the config is valid after creating it. 
It also add the validation at the systemd level, this makes possible to get an error exit code when you restart consul with a faulty configuration. 

Related to https://github.com/ansible-community/ansible-consul/issues/518


templates/consul_systemd.service.j2
```diff
@@ -23,6 +23,7 @@ Environment=CONSUL_UI_LEGACY=true
 {% endif %}
 ExecStartPre=-/bin/mkdir -m 0750 -p {{ consul_run_path }}
 ExecStartPre=/bin/chown -R {{ consul_user }}:{{ consul_group }} {{ consul_run_path }}
+ExecStartPre={{ consul_binary }} validate -quiet {{ consul_config_path }}/config.json {{ consul_configd_path }}
```
abarbare, line 26: You may want to use the same syntax as bellow 

```
{{ consul_bin_path }}/consul
```

tasks/config.yml
```diff
@@ -42,3 +42,6 @@
   set_fact:
     managed_files: "{{ managed_files |default([]) }} + \
       [ '{{ consul_configd_path }}/50custom.json' ]"
+
+- name: Validate configuration
+  shell: "{{ consul_binary }} validate -quiet {{ consul_config_path }}/config.json {{ consul_configd_path }}"
```
nre-ableton:
```suggestion
  command: >
    {{ consul_binary }} validate {{ (ansible_verbosity == 0) | ternary("-quiet", "") }}
    {{ consul_config_path }}/config.json {{ consul_configd_path }}
  changed_when: false
```

Some notes on the above suggestion, since there are several parts to it:


- CI isn't working correctly at the moment (which is something I need to look into), but ansible-lint would [complain about the use of `shell`](https://ansible.readthedocs.io/projects/lint/rules/command-instead-of-shell/).
- Same goes for [`changed-when`](https://ansible.readthedocs.io/projects/lint/rules/no-changed-when/)
- Regarding `-quiet`, the output may be useful because it also contains warnings. Let's only use `-quiet` if no `-v` arguments were given to Ansible.

templates/consul_systemd.service.j2
nre-ableton: Two questions:

1. Why do we need to validate on service startup? We already do it when configuring the task.
2. Why only do this for systemd? There are several more service manager files in `templates` (namely, for SysV init, macOS launchctl, BSD, etc. This change should be applied to all of them, or not at all.

I would suggest dropping this change. I think it's enough to validate the config file after configuring it in `tasks/config.yml`.

nre-ableton (commented):
Thanks for your PR! :+1: I have a few suggestions.

templates/consul_systemd.service.j2
bastienwirtz: @abarbare I made it consistent using `consul_binary` on the other call, so the binary name isn't hardcoded anywhere.

@nre-ableton 
1.  It ensure any external configuration change (manual change for exemple) will be validated. The validate gives you a non 0 return code and system will show you the error right away, instead of failing silently (a systemctl restart consul won't tell you anything about any issue). I thinks it's a nice and free addition, but not mandatory. If you really prefer to remove it I'll do it. The validation during the playbook run is the most important part.  
2. I only have systemd based system to test it, but I can probably use molecule to validated it ? (never done any automated testing with ansible). 

If you are willing to keep the change I can look into adding and testing the change into other service manager init script.

bastienwirtz (commented):
Thanks for your quick review @nre-ableton :pray: !

templates/consul_systemd.service.j2
nre-ableton: > @abarbare I made it consistent using `consul_binary` on the other call, so the binary name isn't hardcoded anywhere.

BTW, I didn't mention this in my last review (since I advocated for removal of that change), but I appreciated use of `consul_binary`. The _other_ code should be changed and we shouldn't use `{{ consul_bin_path }}/consul` anymore, but that's for a future cleanup PR.

> 1. It ensure any external configuration change (manual change for exemple) will be validated. The validate gives you a non 0 return code and system will show you the error right away, instead of failing silently (a systemctl restart consul won't tell you anything about any issue). I thinks it's a nice and free addition, but not mandatory. If you really prefer to remove it I'll do it. The validation during the playbook run is the most important part.
> 
> 2. I only have systemd based system to test it, but I can probably use molecule to validated it ? (never done any automated testing with ansible).
>
> If you are willing to keep the change I can look into adding and testing the change into other service manager init script.

@bastienwirtz I agree that the verification check would be useful to have before starting the service, and you've made a good case for having it here. However, we should try to maintain feature parity and consistency in this role and not get in the habit of adding features for some systems but not others. I did a brief bit of digging, and fully supporting it on all systems would involve the following:

- FreeBSD (`consul_bsdinit.j2`): just a matter of adding the extra command in the `consul_start()` function
- SysV init (`consul_sysvinit.j2`): same thing, except the function is named `start()`
- Debian non-systemd (`consul_debianinit.j2`): we could actually remove this, because systemd is the [default service manager on Debian since version 8 (jessie)](https://wiki.debian.org/systemd), which [isn't supported by this role anymore](https://github.com/ansible-community/ansible-consul/pull/549).
- macOS (`consul_launchctl.plist.j2`): this is where things start to get tricky. AFAIK, launchctl doesn't support anything like `ExecStartPre` in systemd. It *might* be possible to make a secondary service and check the result of that, but I'm not sure that this would work if it were a one-shot service. I don't think that launchctl allows you to do things like `&&` in the `ProgramArguments` array.
- Windows (`tasks/windows.yml`): on Windows, we don't need a template, because [`win_service`](https://docs.ansible.com/ansible/latest/collections/ansible/windows/win_service_module.html) does everything we need. The implementation here might be similar to macOS by using a separate service. I'm also not sure whether `win_service` allows you to have compounded commands joined by `&&`.

I understand that the burden of implementing this on all platforms (especially Mac/Windows) is unfair to ask of you. I would therefore suggest that we compromise with the following approach:

1. Add a separate commit to remove support for Debian non-systemd. This should just be a matter of removing the `consul_debianinit.j2` file, and the corresponding `template` task in `tasks/nix.yml`. We could also even unofficially support it by changing the first condition of the "Create SYSV init script" task to be `ansible_service_mgr == "sysvinit"` instead of the current `ansible_service_mgr != "systemd"` (which is arguably more correct anyway).
2. Add support for FreeBSD and SysV init, as detailed above.
3. We can make an issue in the backlog to provide the feature for Mac/Windows systems. This way, the lack of feature parity is at least documented, and perhaps someone from the community will have a good solution for either (or both) of these systems.

Does that sound like a good approach? Alternately, we can drop support for this in the service start if you think it's too much of a pain.

bastienwirtz: Sounds like a plan @nre-ableton! Thanks for digging into it, I'll work on it, glad to contribute :)

bastienwirtz: Updated @nre-ableton :+1:

bastienwirtz:
@nre-ableton I added a commit to validate the configuration using the consul user. Some config checks can involve the user context (like checking the max open file descriptor allowed, which can be set at the user level (ulimit)).

nre-ableton (commented):
Sorry for the late review, but we've finally got CI working again, so we can proceed with this PR. Anyways, the actual changes look good but I would like to tighten up the git history a bit. It would be nice if there were just 2 commits, "Add support for configuration validation" and "Drop support for non-systemd Debian systems".

So in other words, can you please squash c5fba55e342126cb3be0fbe6b575f0d3a4700fd0 and 9903f5a4582f98c0437117663ba65c183fa41ec2 into ae1c05d69f46e906a3d6509fe98369c7b2cd21c5? Thanks. :pray: 

Also, you'll need to rebase your branch to get it to pass CI, which unfortunately also means fixing a few merge conflicts as well as some ansible-lint violations.

bbaassssiiee (commented):
Nice work!

bastienwirtz:
We should be good here too @nre-ableton

nre-ableton:
Hmm, something is weird with CI. I'm not sure that the merge commit went as intended...

If I have time this afternoon, I'll try to fix it and force-push to your branch. Can you please make sure that this PR has the option ticked to allow contributions from owners?

bastienwirtz:
`edits by maintainers` is enabled, that should be all good.

nre-ableton:
Ok, there are two issues that cause CI to fail:

- On Debian, the task "Check Consul HTTP API (via TCP socket)" fails with the error `Timeout when waiting for 127.0.0.1:8500`. I'm not sure if it was a good idea to remove the `consul_debianinit.j2` file; we may need it around mostly for testing Docker containers.
- On Ubuntu, the task "Start Consul" fails with an error: `Job for consul.service failed because the service did not take the steps required by its unit configuration. See "systemctl status consul.service" and "journalctl -xe" for details.`. The fact that this error message mentions `systemctl` indicates that the Docker container is trying to do `systemd` stuff, which will fail since Docker containers don't support systemd. It should be noted that the Ubuntu molecule tests do correctly detect SysV init, so I'm not sure why this test is failing.

nre-ableton:
@bastienwirtz Could you please rebase this branch and see if CI is still broken (hopefully that's not the case). It would be nice to dust off this PR and get it merged.

nre-ableton:
Bummer, looks like the CI problem is still there. I'll have a look into it today, if I have time.

tasks/config.yml
```diff
@@ -41,3 +41,11 @@
 - name: Set fact list with custom configuration file
   ansible.builtin.set_fact:
     managed_files: "{{ managed_files | default([]) }} + [ '{{ consul_configd_path }}/50custom.json' ]"
+
+- name: Validate configuration
```
rndmh3ro, line 45: Idea (non-blocking): the copy- and template-module have the `validate`-parameter:

```
- validate
        The validation command to run before copying the updated file into the final destination.
        A temporary file path is used to validate, passed in through '%s' which must be present as in the examples below.
        Also, the command is passed securely so shell features such as expansion and pipes will not work.
        For an example on how to handle more complex validation than what this option provides, see Complex configuration validation <https://docs.ansible.com/ansible/devel/reference_appendices/faq.html>.
        default: null
        type: str
```

Maybe we could use this?

nre-ableton: @rndmh3ro We [discussed this](https://github.com/ansible-collections/ansible-consul/pull/565#discussion_r1285584458) earlier in the PR, actually. The [argument](https://github.com/ansible-collections/ansible-consul/pull/565#discussion_r1285802667) that @bastienwirtz made was that validating on system startup would also protect against external changes to the config file made after the playbook was run.

However, I'm also starting to wonder if maybe just validating the config file during templating would be sufficient. We might have a case of "perfect being the enemy of the good" here. @bastienwirtz, what do you think?

nre-ableton:
@bastienwirtz I pushed some fixups to your branch in an (unsuccessful) attempt to fix CI. Feel free to drop them all and force-push.

tasks/config.yml
bastienwirtz: I think @rndmh3ro was talking about the possibility to move the validation command into the `ansible.builtin.copy` which support a `validate` argument (I didn't know, that's cool). it does not relate to the system startup files. It should work, and would be nicer. I can look into it after we fix the issue with the tests. 

@nre-ableton We can remove the startup level validation if you think it helps, but as I understand it, the issue with the tests is related to the fact we decided to clean some unused init system, not that we added a validation command right?

rndmh3ro: @nre-ableton thanks for linking this, I obviously did not see it (I did not check the resolved conversations).

I understand @bastienwirtz' point of view. I also understand your viewpoint for feature-parity in the different OS's.

How about we keep it in the systemd/init-files and move the `Validate configuration` task into the templating/copy-tasks as a `validate`-parameter (this could be done in another PR so this PR here can move forward).

nre-ableton: > I obviously did not see it (I did not check the resolved conversations).

@rndmh3ro No worries, it was well-hidden. 🙂 I agree that perhaps we could break this work up into two PRs, at least having one for the templating task would be nice.

@bastienwirtz To be honest, I'm not sure why the tests are broken. 🙈 I have a hard time running molecule locally for this repo, so it's difficult to tell. I have some bigger plans to refactor this role into a collection, which would facilitate such things, but that's a ways away. In the meantime, I am not sure why this change breaks CI. 😞

bastienwirtz: There you go: https://github.com/ansible-collections/ansible-consul/pull/592
Let me know if that works for you!

bastienwirtz:
replaced by https://github.com/ansible-collections/ansible-consul/pull/592

PR #566, opened 2023-08-07
Add rolling restart on linux.

bastienwirtz:
This patch makes possible to perform rolling restart on linux. When run on an existing cluster, it should keep the cluster up & running.

### DIFF
--- a/README.md
+++ b/README.md
@@ -835,6 +835,14 @@ _Consul Enterprise Only (requires that CONSUL_ENTERPRISE is set to true)_
  - If the default config template does not suit your needs, you can replace it with your own.
  - Default value: `templates/config.json.j2`.
 
+### `consul_rolling_restart`
+ - Restarts consul node one by one to avoid service interruption on existing cluster (Unix platforms only).
+ - Default value: *false*
+
+### `consul_rolling_restart_delay_sec`
+ - Adds a delay between node restart (Linux platforms only).
+ - Default value: 5
+
 #### Custom Configuration Section
 
 As Consul loads the configuration from files and directories in lexical order, typically merging on top of previously parsed configuration files, you may set custom configurations via `consul_config_custom`, which will be expanded into a file named `config_z_custom.json` within your `consul_config_path` which will be loaded after all other configuration by default.
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -36,6 +36,8 @@ consul_checksum_file_url: https://releases.hashicorp.com/consul/{{ consul_versio
 consul_force_install: false
 consul_install_remotely: false
 consul_install_from_repo: false
+consul_rolling_restart: false
+consul_rolling_restart_delay_sec: 5
 
 ### Paths
 consul_bin_path: /usr/local/bin
--- a/handlers/restart_consul.yml
+++ b/handlers/restart_consul.yml
@@ -13,6 +13,20 @@
     # Needed to force SysV service manager on Docker for Molecule tests
     use: "{{ ansible_service_mgr }}"
   when:
+    - not consul_rolling_restart | bool
+    - ansible_os_family != "Darwin"
+    - ansible_os_family != "Windows"
+  listen: restart consul
+
+- name: Rolling restart consul on Unix
+  # Manual loop on playbook hosts to perform a "serial: 1" equivalent on a set of task
+  ansible.builtin.include_tasks: ../tasks/leave_restart_consul.yml
+  with_items: "{{ play_hosts }}"
+  loop_control:
+    loop_var: rolling_restart_host
+  when:
+    - consul_rolling_restart | bool
+    - inventory_hostname == play_hosts[0]
     - ansible_os_family != "Darwin"
     - ansible_os_family != "Windows"
   listen: restart consul
--- a/tasks/leave_restart_consul.yml
+++ b/tasks/leave_restart_consul.yml
@@ -1,0 +1,19 @@
+---
+
+- name: Consul leave
+  delegate_to: "{{ rolling_restart_host }}"
+  ansible.builtin.command: "{{ consul_binary }} leave {% if consul_acl_enable %} -token {{ consul_acl_master_token }} {% endif %} -http-addr {{ consul_addresses.http }}:{{ consul_ports.http }}"
+  changed_when: true
+
+- name: Restart consul on Unix
+  delegate_to: "{{ rolling_restart_host }}"
+  ansible.builtin.service:
+    name: consul
+    state: restarted
+    # Needed to force SysV service manager on Docker for Molecule tests
+    use: "{{ ansible_service_mgr }}"
+
+- name: Wait for service availability
+  ansible.builtin.pause:
+    seconds: "{{ consul_rolling_restart_delay_sec }}"
+  when: consul_rolling_restart_delay_sec > 0
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,5 +10,6 @@
     "files.insertFinalNewline": true,
     "files.trimTrailingWhitespace": true,
     "python.testing.pytestEnabled": true,
-    "python.testing.unittestEnabled": false
+    "python.testing.unittestEnabled": false,
+    "ansible.python.interpreterPath": "/bin/python"
 }


handlers/restart_consul.yml
```diff
@@ -6,12 +6,11 @@
...
 
-- name: Restart consul on Unix
-  service:
-    name: consul
-    state: restarted
-    # Needed to force SysV service manager on Docker for Molecule tests
-    use: "{{ ansible_service_mgr }}"
+- name: Rolling restart consul on Unix
+  # Consul leave will perform a clean stop while adding delay between restarts (combined
+  # with the throttle option) so an existing cluster nodes will stay up & running.
```
estellesoulard: nit: typo/unclear?
```suggestion
  # with the throttle option) so the cluster stays up and running.
```

handlers/restart_consul.yml
```diff
@@ -6,12 +6,11 @@
...
+  throttle: 1
+  shell: "{{ consul_binary }} leave {% if consul_acl_enable %} -token {{ consul_acl_master_token }} {% endif %} -http-addr {{ consul_addresses_http }}:{{ consul_ports_http }} && systemctl restart consul.service"
```
nre-ableton: Unfortunately this PR won't work as presently constructed. Dropping the `service` task in favor of a hardcoded `systemctl restart` command means that this will break on non-systemd Linux hosts. Also, although CI isn't working at the moment (I need to look into fixing that when I have some time), [ansible-lint would complain about the `shell` module](https://ansible.readthedocs.io/projects/lint/rules/command-instead-of-shell/). I guess a suppression would be necessary, since the `&&` requires `shell` and can't be cleanly replaced by `command`.

nre-ableton (commented):
Thanks for your PR! :+1: Unfortunately, we need to re-think some things, because we need to retain the `service` module. Otherwise, this PR will break for non-systemd hosts.

handlers/restart_consul.yml
```diff
@@ -6,13 +6,13 @@
...
-    state: restarted
-    # Needed to force SysV service manager on Docker for Molecule tests
-    use: "{{ ansible_service_mgr }}"
+- name: Rolling restart consul on Unix
+  # Manual loop on playbook hosts to perform a "serial: 1" equivalent on a set of task
+  include_tasks: restart-task.yml
+  with_items: "{{ play_hosts }}"
+  listen: 'restart consul'
   when:
+    - inventory_hostname == play_hosts[0]
```
bastienwirtz, line 29: This replace a `run_once`, which seems not functional in handlers

bastienwirtz (commented):
Hello @nre-ableton, 
Thanks for the review!
I agree regarding the service module. I found another approach to this (inspired by [this](https://serverfault.com/a/967205). It makes possible to use command instead of shell and most importantly, keep the system module. However, it's a bit of a hack, not super dirty but still, I would understand if you don't want to go for it (rolling restart is important for me, I would have to maintain a fork for that, I would love not to have to do that but that's fine!).
The main drawback of this approach is a bunch of `skipped` in the ansible output. However, if one of the restart fail, th playbook stop and the cluster stays up. 

Let me know what you think about it :)

nre-ableton:
> I agree regarding the service module. I found another approach to this (inspired by [this](https://serverfault.com/a/967205). It makes possible to use command instead of shell and most importantly, keep the system module. However, it's a bit of a hack, not super dirty but still, I would understand if you don't want to go for it (rolling restart is important for me, I would have to maintain a fork for that, I would love not to have to do that but that's fine!).

Do you mean this answer here? https://serverfault.com/a/967205/2671 -- that's a pretty clever workaround, using `run_once` with `delegate_to` and `with_items`/`loop`. I wonder if that would work with `block` (I suspect not) or `include_tasks` (maybe, but I think that `include_tasks` doesn't support delegation). Anyways, let me know what you find.

> The main drawback of this approach is a bunch of `skipped` in the ansible output.

Personally, I don't think this is a big problem.

> Let me know what you think about it :)   

Regarding the fork, you might also be able to work around it by using a wrapper role and executing _that_ in serial. I can do some experiments later and see if that would work.

*Edit*: As I suspected, `include_role` also doesn't allow `throttle`. 😞

handlers/restart_consul.yml
```diff
@@ -6,13 +6,13 @@
...
 
-- name: Restart consul on Unix
-  service:
-    name: consul
-    state: restarted
-    # Needed to force SysV service manager on Docker for Molecule tests
-    use: "{{ ansible_service_mgr }}"
+- name: Rolling restart consul on Unix
+  # Manual loop on playbook hosts to perform a "serial: 1" equivalent on a set of task
+  include_tasks: restart-task.yml
```
nre-ableton: Shouldn't this be `leave_restart_consul.yml` (which should be in `tasks`, not `handlers`)?

nre-ableton (commented):
Ok, this seems like a promising approach. I have a question on the specific handler. Also, have you tested that it works properly against a group of hosts?

handlers/restart_consul.yml
bastienwirtz: Indeed! I renamed it and forgot to update the include_tasks. That's fixed.

bastienwirtz:
Hello @nre-ableton, 

fyi, I've been able to successfully test all 3 PRs (merged into one branch) on multiple scenario (bootstrap new cluster, configuration update on existing one, add new nodes to an existing cluster - all running on Ubuntu) on my production env, all went well. 

Let me know if you have any question.

bastienwirtz:
Just added an optional rolling restart delay. After some usage, I find it more reliable with a small delay to make sure there is enough time for the service to be ready, elect a new leader if necessary ...
Let me know what do you think!

bastienwirtz:
Hey @nre-ableton, let me know if I can do anything to move this and the 2 other PR forward. Thx :pray:

nre-ableton:
> Hey @nre-ableton, let me know if I can do anything to move this and the 2 other PR forward. Thx 🙏

Hi @bastienwirtz, sorry for the delay, I've been on holidays. Also, the problem is that CI is broken for this project, which is unfortunately not easy to fix. The root cause of the CI failures is that Molecule no longer accepts this role's name, and until we can get around that I'm hesitant to merge any new changes to the repo.

I need to spend some time to see if there's a workaround (and I haven't found one yet), or possibly to move this role to an Ansible collection (see https://github.com/ansible-collections/ansible-inclusion/discussions/64#discussioncomment-6962013).

bastienwirtz:
I understand, no worries @nre-ableton ! I never used Molecule but I'll try to look into it if I can.

tasks/leave_restart_consul.yml
```diff
@@ -0,0 +1,14 @@
+---
+
+- name: Consul leave
+  delegate_to: "{{ item }}"
+  command: "{{ consul_binary }} leave {% if consul_acl_enable %} -token {{ consul_acl_master_token }} {% endif %} -http-addr {{ consul_addresses_http }}:{{ consul_ports_http }}"
+  changed_when: false
```
nre-ableton: Since this changes system state, I would argue it should be `changed_when: true`.

tasks/leave_restart_consul.yml
```diff
@@ -0,0 +1,14 @@
+---
+
+- name: Consul leave
+  delegate_to: "{{ item }}"
```
nre-ableton:
```suggestion
  delegate_to: "{{ rolling_restart_host }}"
```

See my above suggestion for context. Also, we may need to add a pipe with `default(omit, true)` for compatibility with ansible-core 2.16.x.

handlers/restart_consul.yml
```diff
@@ -6,13 +6,13 @@
...
-- name: Restart consul on Unix
-  service:
-    name: consul
-    state: restarted
-    # Needed to force SysV service manager on Docker for Molecule tests
-    use: "{{ ansible_service_mgr }}"
+- name: Rolling restart consul on Unix
+  # Manual loop on playbook hosts to perform a "serial: 1" equivalent on a set of task
+  include_tasks: ../tasks/leave_restart_consul.yml
+  with_items: "{{ play_hosts }}"
```
nre-ableton, line 24: It seems slightly dangerous to me to use Ansible's default `item`. Namely, if someone is invoking this role from inside of a loop, then we'll end up shadowing that variable.

```suggestion
  with_items: "{{ play_hosts }}"
  loop_control:
    loop_var: rolling_restart_host
```

tasks/leave_restart_consul.yml
```diff
@@ -0,0 +1,14 @@
...
+  command: "{{ consul_binary }} leave {% if consul_acl_enable %} -token {{ consul_acl_master_token }} {% endif %} -http-addr {{ consul_addresses_http }}:{{ consul_ports_http }}"
+  changed_when: false
+
+- name: Restart consul on Unix
+  delegate_to: "{{ item }}"
```
nre-ableton:
```suggestion
  delegate_to: "{{ rolling_restart_host }}"
```

nre-ableton (commented):
CI is now fixed, so we can proceed with this PR. Can you please rebase it and fix the conflicts? Thanks. :+1:

tasks/leave_restart_consul.yml
bastienwirtz: could it be the existing `consul_servers` vars instead of a new one?

tasks/leave_restart_consul.yml
bastienwirtz: I agree :+1: ~I just removed the `changed_when`, as it should be the same as set it to true.~ It's now set to `true`.

handlers/restart_consul.yml
```diff
@@ -6,13 +6,14 @@
...
+  loop_control:
+    loop_var: consul_servers
```
nre-ableton:
```suggestion
    loop_var: consul_server
```

Not to nitpick, but it would be nice if the `loop_var` was singular instead of plural, because it's a bit confusing then in `leave_restart_consul.yml` and looks like the tasks are being delegated to multiple hosts.

tasks/leave_restart_consul.yml
```diff
@@ -0,0 +1,19 @@
+---
+
+- name: Consul leave
+  delegate_to: "{{ consul_servers }}"
+  ansible.builtin.command: "{{ consul_binary }} leave {% if consul_acl_enable %} -token {{ consul_acl_master_token }} {% endif %} -http-addr {{ consul_addresses_http }}:{{ consul_ports_http }}"
```
nre-ableton:
```suggestion
  ansible.builtin.command: "{{ consul_binary }} leave {% if consul_acl_enable %} -token {{ consul_acl_master_token }} {% endif %} -http-addr {{ consul_addresses.http }}:{{ consul_ports.http }}"
```

I _think_ that this should fix the CI error, which is `The task includes an option with an undefined variable. The error was: 'consul_addresses_http' is undefined.`. The port might need to be changed as well.

tasks/leave_restart_consul.yml
```diff
@@ -0,0 +1,19 @@
+---
+
+- name: Consul leave
+  delegate_to: "{{ consul_servers }}"
```
nre-ableton:
```suggestion
  delegate_to: "{{ consul_server }}"
```

....if you agree with my corresponding suggestion in `handlers/restart_consul.yml`

tasks/leave_restart_consul.yml
```diff
@@ -0,0 +1,19 @@
...
+  ansible.builtin.command: "{{ consul_binary }} leave {% if consul_acl_enable %} -token {{ consul_acl_master_token }} {% endif %} -http-addr {{ consul_addresses_http }}:{{ consul_ports_http }}"
+  changed_when: true
+
+- name: Restart consul on Unix
+  delegate_to: "{{ consul_servers }}"
```
nre-ableton:
```suggestion
  delegate_to: "{{ consul_server }}"
```

....if you agree with my corresponding suggestion in `handlers/restart_consul.yml`

handlers/restart_consul.yml
```diff
@@ -6,13 +6,14 @@
...
 
-- name: Restart consul on Unix
-  ansible.builtin.service:
-    name: consul
-    state: restarted
-    # Needed to force SysV service manager on Docker for Molecule tests
-    use: "{{ ansible_service_mgr }}"
+- name: Rolling restart consul on Unix
+  # Manual loop on playbook hosts to perform a "serial: 1" equivalent on a set of task
+  include_tasks: ../tasks/leave_restart_consul.yml
```
nre-ableton:
```suggestion
  ansible.builtin.include_tasks: ../tasks/leave_restart_consul.yml
```

...to fix the `ansible-lint` violation.

nre-ableton (changes requested):
I made some suggestions to fix the linter error and hopefully also CI. 🙂 

Also, I'm wondering if we should make this behavior "opt-in" and not the default. For users who aren't paying attention to the changelog or history and are just getting the latest version, it'll be both a change in behavior and make their deployments take longer (since the 5 second delay is the default). I think we should consider introducing a new `consul_rolling_restart` variable in `defaults/main.yml` that is `false` by default and only enact the rolling restart when it is `true`. What do you think about that @bastienwirtz?

handlers/restart_consul.yml
bastienwirtz: No worries, my bad, `rolling_restart_host` is better, let use it!

bastienwirtz:
@nre-ableton I agree, an opt-in option is better, I added it and took care of other feedback, linting is happy, Molecule is not 😢, not sure why yet. I manually done some test on Ubuntu VMs with and without rolling restart enabled, it worked fine in both case.

nre-ableton:
Something is wrong with molecule unrelated to this (and your other) PR. I'm investigating now but might not have time to push a PR today to fix it.

bastienwirtz:
No rush @nre-ableton, and thanks for looking into it!

nre-ableton:
CI is fixed (again), can you please rebase? 🙏

bastienwirtz:
Damn, almost there. Maybe just a retry on the debian-11 tests @nre-ableton? I don't think I can do it myself

nre-ableton:
> Damn, almost there. Maybe just a retry on the debian-11 tests @nre-ableton? I don't think I can do it myself

It's ok, I got it. Looks like an HTTP 500 error from galaxy.

defaults/main.yml
```diff
@@ -5,6 +5,8 @@
 consul_debug: false
 virtualenv: "{{ lookup('env', 'VIRTUAL_ENV') | default('') }}"
 consul_install_dependencies: true
+consul_rolling_restart: false
+consul_rolling_restart_delay_sec: 5
```
nre-ableton: Let's move these variables to the "Install Method" section (on line 37), I think they make more sense there compared to the "Core" section.

README.md
```diff
@@ -835,6 +835,14 @@ _Consul Enterprise Only (requires that CONSUL_ENTERPRISE is set to true)_
  - If the default config template does not suit your needs, you can replace it with your own.
  - Default value: `templates/config.json.j2`.
 
+### `consul_rolling_restart`
+ - Restarts consul node one by one to avoid service interruption on existing cluster (Linux platforms only).
```
nre-ableton:
```suggestion
 - Restarts consul node one by one to avoid service interruption on existing cluster (Unix platforms only).
```

Since technically it should work on other platforms, and the tasks do state "Unix platforms only".

nre-ableton (changes requested):
Ok, it's looking much better, we're almost there! Just two more tiny suggestions and then I think we're good to go.

bastienwirtz:
Updated @nre-ableton !
Thanks for your work on the CI :pray:

nre-ableton (approved):
Looks good, thanks for your contribution and sorry about the delay!

PR #570, opened 2023-11-04
Updated installs for Debian 12

twimprine:
Debian checks the python environment and will not install the netaddr packages through python - instead it needs to be installed at a package via apt.

Added the checks to keep it from executing in the python env and installing with apt

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -6,6 +6,7 @@
                     from_json).get('tag_name') | replace('v', '') }}"
   when: 'consul_version == "latest"'
 
+
 - name: Install python dependencies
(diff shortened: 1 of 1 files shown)


bastienwirtz:
Hi there, 
It's kind of a duplicate of https://github.com/ansible-community/ansible-consul/pull/564
The package does not necessarily have to be installed with apt, the issue is that the virtualenv is not activated properly (bug aside, ansible-consul role support isolated python env using virtual env, which works well on Debian and other systems).

nre-ableton (commented):
I agree with @bastienwirtz that #564 would be a better solution.

PR #574, opened 2024-01-03
Fix ansible lint

rndmh3ro:


nre-ableton:
Thanks for the PR @rndmh3ro, but unfortunately CI is broken due to the role name (though I see that you've already run into this and tried to fix it). I am working on refactoring this role into a collection and will post a proposal soon.

rndmh3ro:
@nre-ableton, I was able to work-around this. It may not be pretty but we can build upon these changes. So please take a look. :)

Nethertheless converting this into a collection is the right way to go.

nre-ableton:
Yes, I see that you got CI working, though I need to review this code before approving/merging. I agree that the solution isn't pretty, but it might be enough to help get us migrated to the collection. I'd prefer if the molecule fixes in particular were short-term. Nonetheless, I'll review this PR, but I'd prefer to merge it after we get some clarity on the migration process (but if that ends up taking too long, we can always merge anyways).

As promised, here's my comment on the proposed collection migration process: https://github.com/ansible-collections/ansible-inclusion/discussions/64#discussioncomment-6962013

.ansible-lint
```diff
@@ -1,26 +1,18 @@
 ---
+exclude_paths:
+  - .github/
 skip_list:
-  # TODO: Fix ansible-lint violations for these errors and remove from skip_list
```
nre-ableton, line 3: Please keep this comment here. Some of the items in the skip list (such as `no-handler`) will probably remain forever, but others (like `risky-file-permissions`) should indeed be fixed.

.github/workflows/ansible-lint.yml
```diff
@@ -7,7 +7,6 @@ on:  # yamllint disable-line rule:truthy
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [master]
-
```
nre-ableton, line 9: Please keep this newline, unless there is a good reason to get rid of it. The idea was to have separate sections for the name, `on` triggers, and jobs.

.github/workflows/molecule.yml
```diff
@@ -6,7 +6,6 @@ on:
     branches:
       - master
   pull_request:
-
```
nre-ableton: Please keep this newline, unless there is a good reason to get rid of it. The idea was to have separate sections for the name, `on` triggers, and jobs.

.github/workflows/molecule.yml
```diff
@@ -33,9 +33,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Install test requirements
-        run: pip install -r requirements.txt
-
```
nre-ableton: So this is actually not necessary? If so, please drop daec0cebd0a066437b3a882f10b949675e3201e5.

nre-ableton (changes requested):
Thanks for the effort here! :+1: 

Regarding the commit history, it would be nice if you could get rid of the merge commit (234d967da16e138a98e1ac90f0dbaba314835057). Can you just rebase that branch into this branch instead?

Please squash all of the commits needed to get CI working (6080d180240996ca506b3f8577cb3cf4f7a9b279, 07c6af46f0a99d431a4c125aaad115d63c557f5c, https://github.com/ansible-community/ansible-consul/pull/574/commits/8efc4e76fad8cd043f430c3d045338398c11614a, and https://github.com/ansible-community/ansible-consul/pull/574/commits/abc9e82d0f13e737ba0bb9b5af216f0d77280c7b)

Also there's a typo in the commit message of `7867ac78c863dbc74dc5c80752730daf294f688d` ("use proper role name and **namespac**")

.github/workflows/ansible-lint.yml
```diff
@@ -1,46 +1,23 @@
 ---
-name: Ansible-lint
+name: Ansible Lint
 
-on:
+on:  # yamllint disable-line rule:truthy
```
bbaassssiiee: This .yamllint config item takes care of the 'on:', the disable-line is not needed.

 `check-keys: false`

.yamllint
```diff
@@ -10,5 +10,5 @@ rules:
     level: error
   line-length: disable
   truthy:
-    allowed-values: ['true', 'false']
+    allowed-values: ["true", "false"]
```
bbaassssiiee, line 13: Please keep the single quotes for immutable strings

molecule/_shared/converge.yml
```diff
@@ -2,10 +2,10 @@
 - name: Converge
   hosts: all
   roles:
-    - role: ansible-consul
+    - role: brianshumate.consul
```
bbaassssiiee: Brian is no longer involved since 2020

.ansible-lint
bbaassssiiee: You can also annotate the lines with each violations and remove the skip list to avoid new violations creeping in.

bbaassssiiee (commented):
It's great that you're involved in fixing technical debt of this role. I'm not sure if all your changes are fixes of problems reported by (your version of) ansible-lint. It seems that you're changing some formatting as well. This role used to have an (implicit) code-style where strings having Jinja2 were double-quoted, and strings having non-alphabetic characters were single-quoted. Would documenting style choices be an idea?

bbaassssiiee (dismissed):

nre-ableton:
> Would documenting style choices be an idea?

Yes, we should do that at some point. If we adopt a quoting style, it should probably be the same one that Ansible uses.

To that end, Ansible is not very clear about quoting style, the only explicit recommendation I've found on the topic is in the [Windows guide](https://docs.ansible.com/ansible/latest/os_guide/windows_usage.html#yaml-style):

> You should only quote strings when it is absolutely necessary or required by YAML, and then use single quotes.

It would be nice if `ansible-lint` enforced this (see https://github.com/ansible/ansible-lint/discussions/3937), but that probably won't happen for awhile.

.ansible-lint
rndmh3ro: @bbaassssiiee, good idea, I did that instead of globally excluding them.

.github/workflows/ansible-lint.yml
rndmh3ro: I readded it.

.github/workflows/molecule.yml
rndmh3ro: I readded it.

.github/workflows/ansible-lint.yml
rndmh3ro: fixed.

molecule/_shared/converge.yml
rndmh3ro: I changed it to `community` (let's see if it works in github-actions)

rndmh3ro:
>  This role used to have an (implicit) code-style where strings having Jinja2 were double-quoted, and strings having non-alphabetic characters were single-quoted. Would documenting style choices be an idea?

> It would be nice if ansible-lint enforced this (see https://github.com/ansible/ansible-lint/discussions/3937), but that probably won't happen for awhile.

I actually used `ansible-lint --fix` - and it changed the single quotes to double quotes. So I'm in favor of keeping this standard. If ansible-lint decides to use single-quotes or makes it configurable, we can change it IMO.

nre-ableton:
> I actually used `ansible-lint --fix` - and it changed the single quotes to double quotes. So I'm in favor of keeping this standard. If ansible-lint decides to use single-quotes or makes it configurable, we can change it IMO.

Until we have a standard or style guide, I'm personally ok with this approach.

molecule/_shared/base.yml
```diff
@@ -30,10 +30,10 @@ provisioner:
   playbooks:
     prepare: ../_shared/prepare.yml
     converge: ../_shared/converge.yml
+    verify: ../_shared/verify.yml
```
nre-ableton, line 33: Technically, this commit should also contain the addition of the new `verify.yml` file. However, it's added in a later commit so I won't nitpick about that.

molecule/_shared/base.yml
```diff
@@ -30,10 +30,10 @@ provisioner:
...
   inventory:
     group_vars:
       consul_instances:
         consul_node_role: bootstrap
 verifier:
-  name: testinfra
-  directory: ../_shared/tests
```
nre-ableton, line 39: I guess that with this PR, the two Python files are obsolete, right? I noticed that in the previous commit you removed the `flake8` verification, which only would have applied to them. If these files are indeed unused, please remove them.

nre-ableton (commented):
Looking much better! Thanks for squashing the git history. :+1: 

af95e51d2a2a71d753177b2327fa36fead6d7c81 has a typo in the commit message 🙈 ("make **testss** work...").

bbaassssiiee (dismissed):

bbaassssiiee:
Nice to see that this is in good hands now.

molecule/_shared/base.yml
rndmh3ro: Thanks, fixed!

nre-ableton (approved):
Looks great! Very nice work. :+1:

PR #582, opened 2024-01-16
Fix CI

nre-ableton:
Attempting to install `netaddr` in `tasks/main.yml` resulted in an error:

```
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```

Fortunately, we already install `netaddr` in `prepare.yml` and can define `consul_install_dependencies: false`, since we don't require any other dependencies.

This PR also updates the `checkout` and `molecule` GitHub actions to their latest versions.

### DIFF
--- a/.github/workflows/molecule.yml
+++ b/.github/workflows/molecule.yml
@@ -35,7 +35,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Molecule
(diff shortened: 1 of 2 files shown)


bbaassssiiee (approved):
Hurray! CI fixed!

nre-ableton:
@bbaassssiiee thanks for merging!

PR #591, opened 2024-02-22
introduced consul_systemd_service_name variable

isakrubin:
##### SUMMARY
introduced consul_systemd_service_name variable to allow specifying name of service (and related directories), solving the issue of running multiple instances on the same machine

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/tasks/install_linux_repo.yml
+++ b/tasks/install_linux_repo.yml
@@ -22,7 +22,7 @@
 
     - name: Stop service consul, if running
       ansible.builtin.service:
-        name: consul
+        name: "{{ consul_systemd_service_name }}"
         state: stopped
         # Needed to force SysV service manager on Docker for Molecule tests
         use: "{{ ansible_service_mgr }}"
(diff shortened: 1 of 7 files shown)


tasks/install_linux_repo.yml
```diff
@@ -94,7 +94,7 @@
 - name: Override systemd service params
   ansible.builtin.template:
     src: consul_systemd_service.override.j2
-    dest: /etc/systemd/system/consul.service.d/override.conf
+    dest: "/etc/systemd/system/{{ consul_systemd_service_name }}.service.d/override.conf"
```
nre-ableton, line 97: This causes an ansible-lint violation, since templating Ansible task names only works if the template is at the very end of the string. You'll need to reword the task (and drop the quotes) to make this work.

nre-ableton (changes requested):

rndmh3ro:
Can you add the new variable to the readme, please? Otherwise this looks good to me (when ansible-lint is happy).

PR #628, opened 2026-05-04
Fix ansible 12

xael-fry:
### SUMMARY

When running the role with ansible 12, it fails for multiple reasons:
- Conditional result (True) was derived from value of type 'str'
- consul_recursors fails validation as an arrqay is expected instead of a string `"[]"`

###  ISSUE TYPE

    Bugfix Pull Request

### COMPONENT NAME

Task

### ADDITIONAL INFORMATION

The fix :
- ensure condition to be a boolean
- consul_recursors convert to enforce to be an array


MarcWazo:
@bbaassssiiee please Fix for making it usable with ansible 12

bbaassssiiee (approved):
LFTM

